PR #7, opened 2016-10-07
Regenerated bindings from 0.34.x

et1975:
Beaware: breaking changes.

Took TS definitions from DT, reimported and reorganized to make it compile.
Regenerated Helpers DSL and added unifying types to avoid copy-pasting props across DUs.
Moved less used DUs into enclosing modules to avoid them hiding commonly used DU cases.


react-native/Fable.Helpers.ReactNative.fs
```diff
@@ -157,722 +359,1335 @@ module Props =
...
         | BorderRightColor of string
+        | BorderRightWidth of float
+        | BorderStyle of (* TODO StringEnum solid | dotted | dashed *) string
         | BorderTopColor of string
         | BorderTopLeftRadius of float
         | BorderTopRightRadius of float
+        | BorderTopWidth of float
         | Opacity of float
-        | Overflow of string
+        | Overflow of (* TODO StringEnum visible | hidden *) string
```
alfonsogarciacaro: Shall we add these `StringEnum`s (also for BorderStyle) above by hand? Unfortunately, there are many of these all over the place and could be a bit tedious.

react-native/Fable.Helpers.ReactNative.fs
```diff
@@ -157,722 +359,1335 @@ module Props =
...
+        | OnDateChange of Func<DateTime, unit>
+        | TimeZoneOffsetInMinutes of float
+        | Ref of Ref<DatePickerIOS>
+        interface IDatePickerIOSProperties
+
+    [<KeyValueList>]
+    type DrawerLayoutAndroidProperties =
+        | DrawerBackgroundColor of string
+        | DrawerLockMode of DrawerLockMode 
+        | DrawerPosition of obj
```
alfonsogarciacaro: Unfortunately, this is overwriting [my change from yesterday](https://github.com/fable-compiler/fable-react/commit/0ad6306f4e046f426ff4b639f3c20ab7fdf39381), could we recover that?

react-native/Fable.Helpers.ReactNative.fs
```diff
@@ -157,722 +359,1335 @@ module Props =
...
-        | OnShouldStartLoadWithRequest of (unit -> bool)
-        | RenderError of (unit -> ViewStatic)
-        | RenderLoading of (unit -> ViewStatic)
-        | ScrollEnabled of bool
-        | StartInLoadingState of bool
-        | Style of ViewStyle
-        | Url of string
-        interface IWebViewProperties
+    type LayoutAnimationConfig =
+        | Delay of float //REQUIRED!
```
alfonsogarciacaro: These `REQUIRED` comments could go on top of the case with triple slash so they're visible in the autocompletion tool of the IDE.

```fsharp
/// REQUIRED!
| Delay of float
```

et1975:
Dealing with breaking changes w.r.t Styles/Flex/etc:
 The field, constructor or member 'Margin' is not defined (L37,18-L37,24) (Styles.fs)
> The field, constructor or member 'Margin' is not defined (L44,18-L44,24) (Styles.fs)
> The field, constructor or member 'Margin' is not defined (L52,18-L52,24) (Styles.fs)
> The field, constructor or member 'AlignSelf' is not defined (L58,18-L58,27) (Styles.fs)
> The field, constructor or member 'Padding' is not defined (L59,18-L59,25) (Styles.fs)
> The field, constructor or member 'JustifyContent' is not defined (L63,18-L63,32) (Styles.fs)
> The field, constructor or member 'Flex' is not defined (L64,18-L64,22) (Styles.fs)

Just remove DU type name, and it should find the right source for all the cases.

react-native/Fable.Helpers.ReactNative.fs
et1975: That was more of a comment for my future self - I'm not convinced there's actually such a thing when it comes to react props. But I like your idea.

react-native/Fable.Helpers.ReactNative.fs
et1975: Sorry, the TS definitions I pulled didn't have some of those and somehow I missed it reconciling the differences. Amended the commit.

react-native/Fable.Helpers.ReactNative.fs
et1975: Amended the commit, now with 100% more enums!

react-native/Fable.Helpers.ReactNative.fs
```diff
@@ -108,11 +129,320 @@ module Props =
...
+    type TextDecorationStyle = 
+        | Solid | Double | Dotted | Dashed
+
+    [<StringEnum; RequireQualifiedAccess>]
+    type FontStyle = 
+        | Normal | Italic 
+        
+    [<StringEnum; RequireQualifiedAccess>]
+    type WritingDirection =
+        | Auto | Ltr | Rtl
```
forki: can we use full names here and just put Ltr in the generated JS?

react-native/Fable.Helpers.ReactNative.fs
```diff
@@ -108,11 +129,320 @@ module Props =
...
+        | [<CompiledName("500")>] F500 
+        | [<CompiledName("600")>] F600 
+        | [<CompiledName("700")>] F700 
+        | [<CompiledName("800")>] F800 
+        | [<CompiledName("900")>] F900
+
+    [<StringEnum; RequireQualifiedAccess>]
+    type TextDecorationLine =
+        | None | Underline 
+        | ``Line-through``
```
forki: Please use real F# names + CompiledName attribute

react-native/Fable.Helpers.ReactNative.fs
```diff
@@ -137,12 +463,16 @@ module Props =
         | BorderTopWidth of float
         | BorderWidth of float
         | Bottom of float
-        | Flex of int
-        | FlexDirection of string
-        | FlexWrap of string
+        | Flex of float
```
forki: can this be any number?

react-native/Fable.Helpers.ReactNative.fs
```diff
@@ -157,722 +487,1334 @@ module Props =
...
+        | ZIndex of float
         interface IFlexStyle
 
-    [<KeyValueList>]
-    type IViewStyle =
-        interface end
 
     [<KeyValueList>]
     type ViewStyle =
-        // FlexStyle
```
forki: why did you remove this?

react-native/Fable.Helpers.ReactNative.fs
forki: I tried to put required props in the ctor calls.

react-native/Fable.Helpers.ReactNative.fs
et1975: Not sure, but looking at https://github.com/facebook/css-layout/tree/master/csharp/Facebook.CSSLayout the values are floats, which is consistent with other values, like margin.

react-native/Fable.Helpers.ReactNative.fs
et1975: Personally, I'd want to stick to the original RN names because searching for usage help might be easier that way. In this specific case `Rtl` and `Ltr` are common UI design abbreviations.

PR #8, opened 2016-10-08
Initial attempt to map barcode scanner

forki:
See https://github.com/ideacreation/react-native-barcodescanner

Does it make sense?

### DIFF
--- a/react-native-barcodescanner/Fable.Import.ReactNativeBarcodeScanner.fs
+++ b/react-native-barcodescanner/Fable.Import.ReactNativeBarcodeScanner.fs
@@ -1,0 +1,28 @@
+namespace Fable.Import
+
+open System
+open Fable.Core
+open Fable.Import
+open Fable.Import.JS
+open Fable.Import.Browser
+
+module ReactNativeBarcodeScanner =
+
+    [<StringEnum>]
+    type TorchMode =
+    | On
+    | Off
+
+    [<StringEnum>]
+    type CameraType =
+    | Front
+    | Back
+
+
+    and BarcodeScanner =
+        BarcodeScannerStatic
+
+
+    type Globals =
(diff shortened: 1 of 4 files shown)


alfonsogarciacaro:
Looks good to me :+1:

react-native-barcodescanner/Fable.Import.ReactNativeBarcodeScanner.fs
```diff
@@ -0,0 +1,28 @@
...
+    | Front
+    | Back
+
+
+    and BarcodeScanner =
+        BarcodeScannerStatic
+
+
+    type Globals =
+        [<Import("BarcodeScanner", "react-native-barcodescanner")>] static member BarcodeScanner with get(): BarcodeScannerStatic = failwith "JS only" and set(v: BarcodeScannerStatic): unit = failwith "JS only"
```
alfonsogarciacaro: The error may be here. According the [BarcodeScanner usage instructions](https://github.com/ideacreation/react-native-barcodescanner#usage) you must import the **default** member of "react-native-barcodescanner". This is a bit tricky in ES6 imports, check the difference between these two import statements:

```js
// Import **member** BarcodeScanner from 'react-native-barcodescanner'
import { BarcodeScanner } from 'react-native-barcodescanner';
// Fable [<Import("BarcodeScanner", "react-native-barcodescanner")>]

// Import **default** member from 'react-native-barcodescanner'
// (and give it the name 'BarcodeScanner' in current file)
import BarcodeScanner from 'react-native-barcodescanner';
// Fable: [<Import("default", "react-native-barcodescanner")>]
```

So here you probably need to change the attribute to `[<Import("default", "react-native-barcodescanner")>]`.

forki: It's generating 

    import react_native_barcodescanner from "react-native-barcodescanner";

and that doesn't work

forki: scratch that. that works (but is a bit ugly). I had to set flex to make it visible.

PR #9, opened 2016-10-11
Fixed renderscene requiring ReactElement<ViewProperties>

Zeroto:


forki:
that was wrong? @et1975 what do you think?

Zeroto:
Problem is that everything else returns a `ReactElement<obj>` as type (e.g. `view`) causing type errors for this method

et1975:
I remember `ReactElement<obj>` being used all over the place, it should be fine since we don't use `props` types directly anyway. Maybe longer term we can make some use of this type info, but I don't see how at the moment. Is this the only place?

PR #12, opened 2016-10-14
experimental navigation refinements

et1975:
Inferring from JS sources the properties for cardStack, header and utils.

Breaking: all "touchable" constructors now take children.

Also, addressing some missed opportunities for covariance of props.


forki:
Do you have a sample with usage?

et1975:
Yeah, I'll commit it with the fable-elmish.

forki:
cool

PR #14, opened 2016-10-26
Some small fixes

Zeroto:
Moved ITextStyleAndroid to correct type
Added children to scrollView helper


alfonsogarciacaro:
Thanks, @Zeroto! This should be fine, just in case could you have a look, @forki and @et1975?

PR #15, opened 2016-12-01
Remove pojo 

davidtme:
To match https://github.com/facebook/react/commit/b2297ae6c3117ca9b3e15aedaf73d9187eb88e6c

### DIFF
--- a/src/react/Fable.Import.React.fs
+++ b/src/react/Fable.Import.React.fs
@@ -62,7 +62,7 @@ module React =
     and ReactInstance =
         U2<Component<obj, obj>, Element>
 
-    and [<AbstractClass; Import("Component", "react")>] Component<[<Pojo>]'P, [<Pojo>]'S>(props: 'P) =
+    and [<AbstractClass; Import("Component", "react")>] Component<'P, 'S>(props: 'P) =
         [<Emit("$0.props")>]
         member __.props: 'P = jsNative
         [<Emit("$0.props.children")>]


alfonsogarciacaro:
Hmm, this may require some investigation. According to the [issue that generated the PR you linked](https://github.com/facebook/react/issues/7718) they're removing the warning because of false positives, not because they don't care if it's a plain object or not.

This is an important distinction because React may remove the prototype methods of the object (I guess @et1975 had some trouble with that) and then things may fail at runtime (like equality comparison with `.Equals`).

et1975:
`elmish` no longer passes any user models to React as-is, they are always wrapped in a record, which is never treated as a valid F# object from then on. React shallow-mangles those records, but that's fine. For `fable-react` it might be a good idea to  keep the attr, but I can't think of a specific reason - a callback with a previously supplied argument?

alfonsogarciacaro:
It's common in React to pass functions in the props object, and if this functions are in the prototype I'm afraid they may be removed :/

davidtme:
I've checked and react does keep the functions `.Equals` etc but it does mangle it so `instanceof` no longer works.  It only seems to mangle the root prop so maybe I could wrap my record up?

alfonsogarciacaro:
The helpers in fable-react are a very thin layer on top of React. In a sense, `Pojo` forces you to wrap your record. More high-level libraries like fable-elmish can do this automatically for you. I'd rather keep the `Pojo` attribute for now, because it makes the user aware of the semantics and it plays very nicely with React developer tools.

davidtme:
Thanks @et1975 and @alfonsogarciacaro this has really been helpful. It's not safe to remove the 'Pojo' attribute.

PR #22, opened 2017-04-13
Remove (text: string) arg from TextInput

Zaid-Ajaj:
TextInput cannot have child elements and the text should set in the props rather than in the children. According to a user on gitter:
> When I try to make use of TextInput, passing it an arbitrary string, I get "Raw text cannot be used outside of a <Text> tag. Not rendering string: "foo""

He confirmed that using this alternative works:
```fsharp
let textInput (props: ITextInputProperties list) : React.ReactElement  =
        R.from (RN.TextInput) (unbox props) []
```
Which is the same as 
```fsharp
let textInput (props: ITextInputProperties list) : React.ReactElement  =
       createElement(RN.TextInput, props, [])
```


alfonsogarciacaro:
Thanks!

Zaid-Ajaj:
Np :smile:

PR #23, opened 2017-04-13
Added missing svg attribute width

Zaid-Ajaj:


alfonsogarciacaro:
Thanks! Actually we have a `Width` for HTMLAttr, this will shadow it but I guess it's ok because AFAIK nowadays `width` shouldn't be use directly as an HTML attribute.

PR #24, opened 2017-04-13
Feature/fix void elements

MangelMaxime:
I tried several definition for `voidEl` function.

```fs
let inline voidEl (tag: string) (props: IHTMLProp list) : ReactElement = 
    applySpread createEl (tag, keyValueList CaseRules.LowerFirst props) // No props applied
    applySpread createEl (tag, keyValueList CaseRules.LowerFirst props, null) // Runtime error due to null
    applySpread createEl (tag, keyValueList CaseRules.LowerFirst props, []) // This is working
```

Fix #24

### DIFF
--- a/fable-react/Fable.Helpers.React.fs
+++ b/fable-react/Fable.Helpers.React.fs
@@ -594,7 +594,7 @@ let inline domEl (tag: string) (props: IHTMLProp list) (children: ReactElement l
 
 /// Instantiate a DOM React element (void)
 let inline voidEl (tag: string) (props: IHTMLProp list) : ReactElement =
-    applySpread createEl (tag, keyValueList CaseRules.LowerFirst props)
+    applySpread createEl (tag, keyValueList CaseRules.LowerFirst props, [])
 
 /// Instantiate an SVG React element
 let inline svgEl (tag: string) (props: #IProp list) (children: ReactElement list): ReactElement =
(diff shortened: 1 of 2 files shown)


MangelMaxime:
And I confirm that this change generate errors by the compiler.

alfonsogarciacaro:
Maybe something like `let inline area b = domEl "area" b []` would work but this is also fine :+1: I wonder why `applySpread createEl (tag, keyValueList CaseRules.LowerFirst props)` didn't apply the props, I guess I need to fix `applySpread` (this is just a hacky function I wrote _ad hoc_ for the React bindings).

Thanks a lot!

MangelMaxime:
Yes, would be nice to see if the problem come from `applySpread` as often less is better. :)

alfonsogarciacaro:
I had a look, [this is how `applySpread` works](https://github.com/fable-compiler/Fable/blob/fd1f079a6390af93ad5e224544ddc20c89c26d3d/src/dotnet/Fable.Compiler/Replacements.fs#L667-L673), basically it checks the last argument: if it's a list constructor, it applies the list elements as individual arguments (`applySpread fn(foo, [1; 2; 3])` becomes `fn(foo, 1, 2, 3)`), while for other cases it applies the ES2015 spread operator (`applySpread fn(foo, bar)` becomes `fn(foo, ...bar)`).

That's why `props` fails if it becomes the last argument. Unfortunately this is necessary because users can pass list references so we cannot destruct the individual elements at compile time.

MangelMaxime:
Ok :)
If there is a requirement then we can let it like that :)

@alfonsogarciacaro  One question did you publish a new version on npm ?

alfonsogarciacaro:
@MangelMaxime I just did! `npm i fable-react@next`

Issue #33, opened 2017-07-28
Why HTMLProp.Height attribute is commented ?


MangelMaxime:
Why we should not use `Height` attribute in HTML5 ? [Indicated at this lines](https://github.com/fable-compiler/fable-react/blob/master/Fable.Helpers.React.fs#L415)

Also, it's not the only properties conflicting with `CSSProp`, shouldn't we threat all the same way ?

alfonsogarciacaro:
This was a PR some time ago but you're right, I've also wondered the same a couple of times. Probably it's the time to fix it. I can think of the following options:

- Comment out all the HTMLProps conflicting with CSSProps (maybe not a good idea)
- Uncomment it and let people prefix Height with `CSSProp` when necessary (as it's done for Width)
- Uncomment it but move the CSSProps after HTMLProps so `CSSProp.Height` shadows the HTMLProp version.
- Add `RequireQualifiedAccess` attribute to CSSProps (probably not a good idea either as it would break lots of code)
- Any other option?

MangelMaxime:
I would simply uncomment it.
The compiler will be there to help people detect the change and fix it.

Also, we do this change it can also be the right time to add `data-` case to avoid people doing `!!("data-custom", "value")`

alfonsogarciacaro:
Yeah, you're right, but failures because of shadowed members are very tricky because people don't expect them. I always take a while until I figure them out and there are a couple of issues reported because opening `Fable.Import.Browser` shadows the `FSharp.Core.Option` module.

About `data-` maybe you could do something like this:

```fsharp
let inline Data(key: string, value: obj): IHTMLProp = !!("data-" + key, value)
```

MangelMaxime:
Ah yes didn't think about adding this function to my project.

And I understand for the shadowed members, it cool also be a good time to think about improving the user Experience for SVG.
Example of code:

```fs
let root model dispatch =

    svg [ Style [ CSSProp.Width "100px"
                  Height "100px" ] :> IProp
          ViewBox "0 0 42 42" :> IProp
          ClassName "marker" :> IProp ]
        [ circle [ Cx !^21.
                   Cy !^21.
                   R !^RADIUS
                   Fill "#fff" ]
            [ ]
          circle [ Cx !^21.
                   Cy !^21.
                   R !^RADIUS
                   Fill "transparent"
                   Stroke "#d2d3d4"
                   StrokeWidth !^3. ]
            [ ]
          image [ X !^"50%" :> IProp
[...]

PR #35, opened 2017-08-03
Some API improvements

alfonsogarciacaro:
Also related to #33 

@MangelMaxime please review, I've renamed Width in HTML and SVG props to avoid conflicts, but maybe this is more confusion for users, what do you think?

To my surprise, the problem with mixed properties in SVG elements has been fixed just by removing the `#` in `svgEl` signature. You shouldn't have to cast to `IProp` anymore :)

The last fix we could do is just replace all the `U2<float, string>` with just `obj`. We'll lose a bit of type safety but we don't need to use the `!^` operator constantly (which @forki and I guess others don't like that much :wink:) and ints would also be accepted.

### DIFF
--- a/Fable.Helpers.React.fs
+++ b/Fable.Helpers.React.fs
@@ -603,7 +603,7 @@ let inline voidEl (tag: string) (props: IHTMLProp list) : ReactElement =
     applySpread createEl (tag, keyValueList CaseRules.LowerFirst props, [])
 
 /// Instantiate an SVG React element
-let inline svgEl (tag: string) (props: #IProp list) (children: ReactElement list): ReactElement =
+let inline svgEl (tag: string) (props: IProp list) (children: ReactElement list): ReactElement =
     applySpread createEl (tag, keyValueList CaseRules.LowerFirst props, children)
 
 // Standard element
--- a/Fable.Helpers.React.fs
+++ b/Fable.Helpers.React.fs
@@ -12,6 +12,7 @@ module Props =
         | BoxFlex of float
         | BoxFlexGroup of float
         | ColumnCount of float
+        | Cursor of string
         | Flex of U2<float, string>
         | FlexGrow of float
         | FlexShrink of float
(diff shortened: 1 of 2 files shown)


MangelMaxime:
I think using `obj` instead of `U2<float, string>` isn't so bad. I don't think type checking that important here as we already use `obj` for the inline style which is much more frequent that people using SVG.

One equivalent exemple [here](https://github.com/fable-compiler/fable-react/blob/1ec1ba14ef813f4d109c847bed539923bb58b7d8/Fable.Helpers.React.fs#L279)

I don't think prefixing width and height with `HTML` or `SVG` are really a good benefit. User will not find them easily and also this sounds equivalent to `HTMLAttr.Width` for me.

One solution could be to declare the CSS properties after the HTML and SVG properties. Like so when doing `Width` it's the CSS one being used by default. But this change need to be tested into several production/big apps like mine, @forki ones, etc. to see the real impact. Could be much worse than just adding `CSSProp.Width` depending on the names collision.

inosik:
> just replace all the U2<float, string> with just obj.

I'm not sure if this is a good idea. Using `obj` "forces" me to compile and try the app, or to look up the documentation, to find out if I gave the API the kind of object it expects.

`U2<float, string>` isn't perfect, too, but it's better than nothing.

MangelMaxime:
@inosik I understand but as one of the type is string you can more or less pass anything false value by default on it and so need the runtime to check it.

Extreme example:
`div [ Width "I am a super valid string" ] [ ]`

Also, my idea by making everything an `obj` is to have a consistent API between `HTMLAttr` and `CSSProps`

@inosik I have some ideas to avoid this kind of problem, but I need time to think about it and prototype a library.

@alfonsogarciacaro I didn't saw that you pushed more commits on this PR last time. I will review it and test it.

MangelMaxime:
Sorry for the small commit, I needed it to make Github happy. Was complaning about conflicts..

@alfonsogarciacaro For me it's seems ok. I got some few compilation errors but the compiler was here to help me fixing them.

Would be nice if @forki could take a look when he have time. And perhaps we could release a version 1.2.0-alpha just to be sure as we are doing small breaking changes here.

PR #39, opened 2017-10-25
Rework architecture and add ReactLeaflet

MangelMaxime:
@forki 

So, this PR rework the architecture of the repo to host several packages. For now, there is `Fable.React` and `Fable.ReactLeaflet`.

You can build the package with: `build.cmd Build`
And publish them with: `build.cmd PublishNugets`

This will detect the version from the `RELEASE_NOTES.md` and push only the packages which need to be updated and then update the version in the fsproj when succeeded.

I already publish a first version of `Fable.ReactLeaflet` to test it and it's working.

As discuss on the gitter channel, I am not sure that creating several `Props` DU is the way to go because, some of them need to inherit props from others. So the intellisense isn't helping much here.

Several solutions I can think of:
- Keep separate DU, and let the users know that `Polygon` inherit from `Polyline`. For this, we will need to ensure the inheritance by using interface (not done in the current state).
- Make a unique DU, like for the html/css and let the user use what he want according to the documentation
- Duplicate the inherited case


alfonsogarciacaro:
Great @MangelMaxime, thanks a lot for this! About using DUs to represent JS options: yes, it's tricky when there're inheritance relationships. I did it for React because the syntax with DUs is nicer, but they're giving some issues already for example with the Recharts bindings (see [how an alias is needed to avoid collisions](https://github.com/fable-compiler/fableconf-workshops/blob/1cf3cb5c2f888460086652d68a1358088b822d24/data/src/Recharts/App.fs#L9)). The most flexible option is to just use interfaces for everything which can easily inherit from one another (even with multiple inheritance) and then `jsOptions` to instantiate the options.

MangelMaxime:
Ah yes, I am already using alias module. Problems is reallt with the DU inheritance here.

PR #42, opened 2017-11-04
SVGAttr attribute Height added to match Width

daz10000:
Unless I'm missing something, it's hard to make an SVG rectangle without a Height tag.  There is a matching Width attribute.  Also useful for specifying the SVG element's size as well.  Apologies if I'm missing something deeper.


MangelMaxime:
Thanks for the PR :)

PR #49, opened 2017-11-18
New uniform names for helpers

alfonsogarciacaro:
At the beginning I wrote the helper names to be as short as possible (`com` to instantiate a component from a type, `fn` from a function, `str` from string, etc). But I think this is confusing for new users and React allows more types to be accepted as elements or as the return value of `render()` (numbers, arrays...).

We're going to introduce some breaking changes in 1.3 (now you'll have to use `override` for the `render()` method) so maybe it's good time to try to uniform the helpers API. I was thinking to use more meaningful names prefixed by `of`. The old names would still be valid but they would be marked as obsolete so users can update them. What do you think? @forki @MangelMaxime @et1975 @zaaack 

```
ofType
ofFunction
ofImported
ofString
ofOption
ofInt
ofArray
...
```

### DIFF
--- a/src/Fable.React/Fable.Helpers.React.fs
+++ b/src/Fable.React/Fable.Helpers.React.fs
@@ -583,22 +583,64 @@ module Props =
 
 open Props
 open Fable.Import.React
-open Fable.Core.JsInterop
 
 [<Import("createElement", from="react")>]
-let createElement(componentClass: obj, props: obj, [<ParamList>] children: obj) = jsNative
+let createElement(component: obj, props: obj, [<ParamList>] children: obj) = jsNative
 
-/// Instantiate a React component from a type inheriting React.Component<>
+/// OBSOLETE: Use `ofType`
+[<System.Obsolete("Use ofType")>]
 let inline com<'T,[<Pojo>]'P,[<Pojo>]'S when 'T :> Component<'P,'S>> (props: 'P) (children: ReactElement list): ReactElement =
     createElement(typedefof<'T>, props, children)
 
-/// Instantiate a stateless component from a function
+/// OBSOLETE: Use `ofFunction`
+[<System.Obsolete("Use ofFunction")>]
 let inline fn<[<Pojo>]'P> (f: 'P -> ReactElement) (props: 'P) (children: ReactElement list): ReactElement =
     createElement(f, props, children)
 
 /// Instantiate an imported React component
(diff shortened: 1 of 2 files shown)


MangelMaxime:
Seems a good idea for most of them except for `ofString`/`str`
We are using it literally everywhere in the applications, and I am not sure if we won't introduce a lot of noise in the user code.

Perhaps, I am not used to it but reading `div [ ] [ ofString "I am a div" ]` feels wrong to me.

zaaack:
I'm also thinking the same thing about `ofFunction/ofOption`. @alfonsogarciacaro @MangelMaxime  What about `ofStr/ofFn/ofOpt`?

And one more question, since most option value were wrapped by `Some<T>` class now, should this line
```sh
let inline ofOption (o: ReactElement option): ReactElement = unbox o
```
change to 
```sh
let inline ofOption (o: ReactElement option): ReactElement = 
      match o with 
      | Some v -> v
      | None -> Unchecked.defaultof<ReactElement>
```
?  

[repl link](http://fable.io/repl/#?code=%0Alet%20inline%20ofOption%20(o%3A%20int%20option)%3A%20int%20%3D%20unbox%20o%0A%0A%0Alet%20inline%20ofOption2%20(o%3A%20int%20option)%3A%20int%20%3D%20%0A%20%20%20%20%20%20match%20o%20with%20%0A%20%20%20%20%20%20%7C%20Some%20v%20-%3E%20v%0A%20%20%20%20%20%20%7C%20None%20-%3E%20Unchecked.defaultof%3Cint%3E%0A%0Alet%20makeSome%20a%20%3D%0A%20%20%20%20Some%20a%0A%20%20%20%20%0AofOption%20(makeSome%201)%20%7C%3E%20printfn%20%22%25A%22%0AofOption2%20(makeSome%201)%20%7C%3E%20printfn%20%22%25A%22)

MangelMaxime:
`ofStr/ofFn/ofOpt` isn't easy to read. I would keep `str` because this is the most use one.

For the others, the long version is good.

zaaack:
@MangelMaxime You are right, sorry for my disturb : )

MangelMaxime:
No worries :) Just giving my opinion :)

alfonsogarciacaro:
Thanks a lot for the feedback. OK, we'll keep `str` as alias of `ofString` :+1: And you're right @zaaack, options are still erased in most cases but better be safe than sorry, I'll add the option check :smile:

alfonsogarciacaro:
Fable.React 1.3.0-beta-001 is released, enjoy!

PR #50, opened 2017-12-25
Add BoxShadow CSSProp

worldbeater:
When I tried to write inline style for box-shadow css property, I found out that it is missing in CSSProp discriminated union. So here is a fix.


alfonsogarciacaro:
Great, thanks!

alfonsogarciacaro:
Released as 2.0.0-beta-002 :+1:

PR #54, opened 2018-01-23
Add React PureComponent and Fragment

vbfox:
* Add PureComponent doing shallow state & props comparison to avoid `render()` being called.
* Also add a PureStatelessComponent alias to PureComponent taking no state type parameter. Such components are mapping perfectly to an elm-like architecture with immutable props and no mutable state.
* Add `Fragment` (class) and `fragment` helper function to easily return lists from `render()`


alfonsogarciacaro:
Awesome, thank you!

PR #55, opened 2018-01-27
Added many CSS props sourced from MDN. Changed all CSS types to obj.

chadunit:
(Not sure if you'll want this per issues described below, but submitting at least for consideration/discussion.)

I noticed some fairly typical CSS properties like background-position/background-size weren't represented. Rather than spot-add them, I went ahead and scraped https://developer.mozilla.org/en-US/docs/Web/CSS/Reference and https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute for a more complete list.

Many are experimental/draft settings, but the original list already had some of those so I just left everything in. I'm not sure if the library wants to be in the business of judging what to include/exclude, or leave it to the user.

Note, I didn't test everything added, just a few I was interested in. I converted all hyphen separators to camelCase/PascalCase assuming that works universally, but if that assumption is not true then I guess there might be some invalid entries in that case.

I also converted the small number of float, string, and U2<float,string> types to obj. The floats prevent using the standard CSS "initial"/"unset"/etc. string values, and for the other types I'm not sure it makes sense with regard to consistency to have just a few arbitrary typed properties when everything else is obj. But it is a minor breaking change that might affect some users.

Finally I sorted the entries (so don't be alarmed at block of deletions in the file diff).

### DIFF
--- a/src/Fable.React/Fable.Helpers.React.fs
+++ b/src/Fable.React/Fable.Helpers.React.fs
@@ -312,56 +312,71 @@ module Props =
         interface IProp
 
     type CSSProp =
-        | BoxFlex of float
-        | BoxFlexGroup of float
-        | ColumnCount of float
-        | Cursor of string
-        | Flex of obj
-        | FlexGrow of float
-        | FlexShrink of float
-        | FontWeight of obj
-        | LineClamp of float
-        | LineHeight of obj
-        | Opacity of float
-        | Order of float
-        | Orphans of float
-        | Widows of float
-        | ZIndex of float
-        | Zoom of float
-        | FontSize of obj
-        | FillOpacity of float
-        | StrokeOpacity of float
-        | StrokeWidth of float
         | AlignContent of obj
         | AlignItems of obj
         | AlignSelf of obj
         | AlignmentAdjust of obj
         | AlignmentBaseline of obj
+        | All of obj
+        | Animation of obj
         | AnimationDelay of obj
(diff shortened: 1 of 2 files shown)


alfonsogarciacaro:
Thanks a lot for this @chadunit! Originally I created the CSSProp union semiautomatically from the React TypeScript declaration and we've added some more later manually. But I think it makes sense to add all possible attributes. About the naming, according to [React docs](https://reactjs.org/docs/dom-elements.html#style) this should be fine (Fable will just change the first letter to lower case).

I also agree that it's better to use `obj` everywhere for consistency, it shouldn't break old code (unless you used the `!^` operator). So in principle I'm OK to merge this. Maybe other contributors like @MangelMaxime or @zaaack can have a look too just in case?

MangelMaxime:
This seems ok for me. :)

MangelMaxime:
@alfonsogarciacaro I just think of something.
Should we release this changes under `3.0.0` or `2.2.0` ?

Because, they introduce a lot of signature change, I am in favor of 3.0.0

alfonsogarciacaro:
I agree :+1: Are we merging ICSSProp/CSSProp before that?

MangelMaxime:
I am ok for the merge.

alfonsogarciacaro:
Do you want to send another PR for that @chadunit?

chadunit:
@alfonsogarciacaro Ok I will.

PR #58, opened 2018-02-05
Removed ICSSProp to reduce casting (fix #56)

### LINKED ISSUE #56: ICSSProp vs. CSSProp?
I'm curious what is the use for ICSSProp? I have some hassles using ICSSProp that work better using CSSProp directly.

Consider this initial (working) code:
```
let view1 model =
  div [
    Style [
      BackgroundColor model.BackgroundColor
      BackgroundImage model.BackgroundImage
      BackgroundPosition model.BackgroundPosition
    ]
  ] []
```

Now suppose I want to use yield syntax to have conditional style properties:
```
let view2 model =
  div [
    Style [
      yield BackgroundColor model.BackgroundColor
      if model.HasBackgroundImage then
        yield BackgroundImage model.BackgroundImage
        yield BackgroundPosition model.BackgroundPosition
    ]
  ] []
```

That won't compile:
```
This expression was expected to have type
    'ICSSProp'    
but here has type
    'CSSProp'
```

So I have to cast or unbox everything:
```
let view3 model =
  div [
    Style [
      yield BackgroundColor model.BackgroundColor :> ICSSProp // cast example
      if model.HasBackgroundImage then
        yield !! BackgroundImage model.BackgroundImage // unbox examples
        yield unbox BackgroundPosition model.BackgroundPosition
    ]
  ] []
```

However, if I define my own `Style'` to use CSSProp instead of ICSSProp, then I don't have the extra hassle of casting/unboxing and it still works fine at runtime:
```
let inline Style' (css: CSSProp list): HTMLAttr =
    !!("style", keyValueList CaseRules.LowerFirst css)

let view4 model =
  div [
    Style' [
      yield BackgroundColor model.BackgroundColor
      if model.HasBackgroundImage then
        yield BackgroundImage model.BackgroundImage
        yield BackgroundPosition model.BackgroundPosition
    ]
  ] []
```

Unless there's something important about ICSSProp or some better syntax I can use, maybe the internal `Style` could be defined this way?

chadunit:
Removed ICSSProp as it was only implemented by CSSProp and therefore only added extra casting for no gain. Style helper now takes CSSProp directly. Any users that were explicitly referencing ICSSProp type will need to change. Fixes #56.

### DIFF
--- a/src/Fable.React/Fable.Helpers.React.fs
+++ b/src/Fable.React/Fable.Helpers.React.fs
@@ -5,9 +5,6 @@ open Fable.Core.JsInterop
 open Fable.Import
 
 module Props =
-    type ICSSProp =
-        interface end
-
     type IProp =
         interface end
 
@@ -244,12 +241,6 @@ module Props =
         | Security of string
         | Unselectable of bool
         interface IHTMLProp
-
-    let inline Style (css: ICSSProp list): HTMLAttr =
-        !!("style", keyValueList CaseRules.LowerFirst css)
-
-    let inline Data(key: string, value: obj): IHTMLProp =
-        !!("data-" + key, value)
 
     type SVGAttr =
         | ClipPath of string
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Thanks!

Issue #60, opened 2018-02-23
How to import react elements as members of other elements


Zaid-Ajaj:
Concretely, suppose you have the following in javascript (which seems quite common):
```js
import { Select } from 'react-select' 
let render = () => <Select.Creatable options={...} /> 
```
How to import the `Creatable` element from the `Select` component?

alfonsogarciacaro:
It's funny that you asked that, because in Fable 1.x you can use a `.` to select a member when importing a value, like:

```fsharp
ofImport "Select.Creatable" "react-select" myOptions []
```

...but exactly right now I just removed it from Fable 2 because I thought nobody was using it 😅Can you give it a try and check if it works for you? If it does, I guess I should put it back in dev2.0 :wink:

Zaid-Ajaj:
I tried just now it but it didn't work with my example and I am not sure why. I have this function:
```fs
let creatable (props: ICreatableProps list) = 
    ofImport "Select.Creatable"
             "react-select"
             (keyValueList CaseRules.LowerFirst props)
             [] 
```
I get runtime error: `Cannot read property Creatable of undefined` and also a compile warning `export 'Select' was not found in 'react-select'` which is weird because I have installed `react-select` correctly. Now even if I change `Select.Creatable` to `Select` I get the same warning . . . strange. 

Anyways, if you are thinking of removing this pattern completely, their should be alternative way to do it because major UI frameworks (Elemental/Material/Semantic...) use this nesting to group elements. 

I guess the reason no one is using this feature is because no one has tried yet to port these large frameworks.

alfonsogarciacaro:
This is the usual confusion with JS imports :wink: According to react-select README, it must be imported as follows:

```js
import Select from 'react-select';
```

This is a **default** import (not a **member** import like `import { Select } from 'react-select'`). For this, in Fable you just have to write `"default"`. Can you try the following?

```fsharp
let creatable (props: ICreatableProps list) = 
    ofImport "default.Creatable" "react-select"
             (keyValueList CaseRules.LowerFirst props)
             [] 
```

Zaid-Ajaj:
Wow *facepalm* hahah, this worked like a charm! thanks a lot :heart:

alfonsogarciacaro:
Don't worry, I've made the same mistake thousands of times. Unfortunately, it's not something Fable can check at compile time :/ We need to make it more prominent in the docs.

Thanks for confirming it's working! I'll put back the feature in Fable 2 :+1:

[...]

PR #65, opened 2018-02-27
Add Server Side Rendering Support

zaaack:
### Pure F#, No Nodejs Required!

Recently I'm working on a web app which needs SEO, while search engines in my country don't support SPA well, so server-side-rendering becomes my first choice. At first, I was looking at something like [aspnet/JavaScriptServices](https://github.com/aspnet/JavaScriptServices), but it's a little complicated, and running nodejs instances in an asp.net server would require more overhead. it's just too complicated for my case.

This approach is inspired by Giraffe's view engine, it runs fable-react in the server side by simply replace each HTML tag to an `HTMLNode` tree, so it can be rendered to HTML.

### Props
* Simplier, don't need to install nodejs or maintain nodejs instances on the server-side.
* Easier, everything is F#, easy integration with Elmish.
* Faster, about 2x ~ 3x faster then node + `react-dom/server` (with NODE_ENV=production) in my macbook
* Support React 16's new server-side rendering API (.hydrate)

### Cons
Elements created from js native won't render.

#### Elements support server-side rendering:
* [x] Pure functions (e.g. Elmish Components).
* [x] Class Components written in F# (ofType)
* [x] Function Componets written in F# (ofFunction)
* [ ] React Components from FFI (ofImport)
...


src/Fable.React/Fable.Helpers.ReactServer.fs
```diff
@@ -0,0 +1,688 @@
...
+
+  match htmlNode :?> HTMLNode with
+  | HTMLNode.Text str -> str
+  | HTMLNode.Node (tag, attrs, children) ->
+    let attrs, child = renderAttrs (attrs |> Seq.cast)
+    let child =
+      match child with
+      | Some c -> c
+      | None -> (renderList (children |> Seq.cast))
+    sprintf "<%s %s>%s</%s>" tag attrs child tag
```
alfonsogarciacaro: `sprintf` is a bit slow, I'd use `+` for string concatenation here

Samples/SSRSample/src/Server/Server.fs
```diff
@@ -0,0 +1,82 @@
...
+let htmlTemplate =
+  html []
+    [ head [] []
+      body []
+        [ div [_id "elmish-app"] [ rawText (Fable.Helpers.ReactServer.renderToString(Client.View.view initState ignore)) ]
+          script []
+            [ rawText (sprintf """
+            var __INIT_STATE__ = %s
+            """ (toJson initState)) ]
+          script [ _src "http://localhost:8080/public/bundle.js" ] []
```
alfonsogarciacaro: Even if this is a sample, I'd try to avoid the URL literal here. If you can't use `public/bundle.js` because Wepback dev server and Giraffe are running in different ports, I'd use a variable so to minimize the chances `http://localhost:8080` reaches production.

src/Fable.React/Fable.Helpers.React.fs
```diff
@@ -22,9 +22,13 @@ module Props =
         | Key of string
         | Ref of (Browser.Element->unit)
         interface IHTMLProp
+    [<Pojo>]
+    type DangeousHtml = {
```
alfonsogarciacaro: Minor typo: `DangerousHtml`

zaaack:
@alfonsogarciacaro Sorry for my late,  I was planning to reply in this morning but got a busy day, then I forgot...:sweat_smile:  

I haven't tried this with React 16's new server-side rendering API [.hydrate](https://devdocs.io/react/react-dom#hydrate) yet, but I suppose it will be OK because the HTML generated in server looks really like those from react-dom/server. I'll add it to samples to it works or not.

src/Fable.React/Fable.Helpers.React.fs
```diff
@@ -755,22 +804,21 @@ let inline ofFunction<[<Pojo>]'P> (f: 'P -> ReactElement) (props: 'P) (children:
 /// Example: `ofImport "Map" "leaflet" { x = 10; y = 50 } []`
 let inline ofImport<[<Pojo>]'P> (importMember: string) (importPath: string) (props: 'P) (children: ReactElement list): ReactElement =
```
alfonsogarciacaro: I was thinking to surround `ofImport`, `ofType`, `ofFunction` with  `#if FABLE_COMPILER` to remove them when compiling on server-side. But this would cause conflicts with IDEs on frontend side, because they don't define `FABLE_COMPILER`.  Not sure, what's the best solution. I guess `ofImport` throws an exception quickly (and users won't probably import a JS file in .NET either) but I don't know what happens with `ofType` and `ofFunction`. Do they throw exception or fail silently? 

The other option would be to extract common parts into a `Fable.React.Core` package and distribute two packages: `Fable.React` for JS side and `Fable.ReactServer` or `Fable.React.Net` 🤔

alfonsogarciacaro: We may need a design guideline for shared libraries between JS and .NET. @MangelMaxime is already working on this: https://github.com/haf/Http.fs/issues/142

zaaack: > I guess ofImport throws an exception quickly (and users won't probably import a JS file in .NET either) but I don't know what happens with ofType and ofFunction. Do they throw exception or fail silently?

I was thinking this problem, too. In my imagination, we can create an help function by wrapping runtime exceptions in a function that only executed in the client side, but not tested yet.

```F#
let hybridView (clientView: 'cilentMode -> ReactElement) (clientModel: 'cilentModel) (serverView: 'serverModel -> ReactElement) (serverModel: 'serverModel) =
#if FABLE_COMPILER
    clientView clientModel
#else
    serverView serverModel
#endif
###

Comment (zaaack) [reply src/Fable.React/Fable.Helpers.React.fs]: And import a js module could write like this:


```F#

#if FABLE_COMPILER
let JsComp: React.ComponentClass<JsCompProps> = importDefault "./jsComp"
#else
let JsComp = Unchecked.defaultof<React.ComponentClass<JsCompProps>>
#endif

```

zaaack:
@alfonsogarciacaro `ofType` and `ofFunction` are working now! By using dotnet's reflection API we can render React Components write in F# like react-dom/server!

But I got some trouble when implement a isomorphic version of `import`, any thoughts or advice about this?

[repl](http://fable.io/repl/#?code=open%20Fable.Core%0Aopen%20Fable.Core.JsInterop%0A%0A%5B%3CEmit(%22%22%22%20require('%241').%242%20%22%22%22)%3E%5D%0Alet%20importOrDefault%3C'T%3E%20selector%20path%20%3D%0A%20%20%20%20Unchecked.defaultof%3C'T%3E%0A%5B%3CEmit(%22%240%20%2B%20%241%22)%3E%5D%0A%0Alet%20add%20(x%3A%20int)%20(y%3A%20string)%3A%20float%20%3D%20Unchecked.defaultof%3C'T%3E%0A%0A%0Aadd%201%20%22aa%22&html=%3Chtml%3E%3Cbody%3E%3C%2Fbody%3E%3C%2Fhtml%3E%0A), any thoughts?

```F#
[<Emit(""" require($1)[$2] """)>]
let importOrDefault<'T> (selector: string) (path: string) =
    Unchecked.defaultof<'T>
```

alfonsogarciacaro:
I wouldn't use the `Emit` attribute with import expressions for several reasons:

- JS tree shaking works much better when using `import` instead of `require.
- Fable also does a few tricks with imports (e.g. see https://github.com/fable-compiler/fable-react/issues/60#issuecomment-368182485)

If you've made `ofType` and `ofFunction` work that's awesome! I wouldn't worry too much about `ofImport`: if users try to compile it on the server side, a "JS only" exception will be thrown right away so hopefully users will spot the problem quickly.

I'm OK with merging this :+1: Is it good to go or do you want to work a bit more on it?

zaaack:
@alfonsogarciacaro I think I'm done here if I don't need to implement isomorphic `import`. If I have missed something, just tell me, please.

alfonsogarciacaro:
Awesome, thanks a lot for this amazing feature @zaaack! I'll merge, try locally and let you know if I have any question 😄

forki:
Next step: tutorial how to do it. Please please please

Am 01.03.2018 14:55 schrieb "Alfonso Garcia-Caro" <notifications@github.com
>:

Merged #65 <https://github.com/fable-compiler/fable-react/pull/65>.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/fable-compiler/fable-react/pull/65#event-1499045848>,
or mute the thread

forki:
@zaaack any chance you cou try to add SSR to https://github.com/SAFE-Stack/SAFE-BookStore

also the tutotial should contain some guidance on when to use it and when not. 
Also needs more DotNet core advertising - I know MS loves that

zaaack:
@forki 
> any chance you cou try to add SSR to https://github.com/SAFE-Stack/SAFE-BookStore
The sample project actually is based on the SAFE-Stack template! I think I can send a pr to SAFE-BookStore if it's OK to have SSR as default in SAFE-Stack.


> also the tutotial should contain some guidance on when to use it and when not.
Also needs more DotNet core advertising - I know MS loves that

Good advice, I think I can add some guidance in the tutorial, but I'm not sure how to add DotNet core advertising? Actually I'm not an experienced dotnet developer, not sure about how to do this right...

forki:
that dotnet core advertising is something that we need to take to the MS guys. Don't worry about that. 
You did fantastic work here!

PR #69, opened 2018-03-02
Update server-side-rendering.md

forki:
/cc @zaaack


forki:
image

should we add a bit of CSS?

PR #71, opened 2018-03-04
Add multi-thread benchmark for SSR

zaaack:
As @forki said on Twitter: https://twitter.com/sforkmann/status/969912930501300224 

I added a multi-thread version benchmark in SSRSample, this time I tested in a more clean environment, with all applications that might increase system's load like Chrome, VSCode closed (which I really should do at the first benchmark 😅 ). 

The good news is that fable-react on dotnet core is still faster then `ReactDOMServer` on nodejs, and even faster. In my macbook (2.7 GHz Intel Core i5, 4 processors), it's about ~4x faster in a single thread, and still about ~3x faster in 4 threads/workers!

### DIFF
--- a/Samples/SSRSample/benchmark/dotnet.fs
+++ b/Samples/SSRSample/benchmark/dotnet.fs
@@ -1,0 +1,57 @@
+
+open System
+open System.IO
+open System.Threading
+open System.Threading.Tasks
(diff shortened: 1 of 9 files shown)


forki:
Hugs

PR #73, opened 2018-03-04
Unify Fable.Helpers.React.Props

alfonsogarciacaro:
This is a continuation of discussion in #56. As we removed there the `ICSSProp` interface to reduce casting when using list comprehension, it may be a good opportunity to unify other props. I didn't have much time to think about this, but it may be better to do it before releasing Fable.React 3 (there's already an alpha version out there though) with SSR so users don't have to change a lot of code later. I just quickly edited `Fable.Helpers.React.fs` but didn't fix the other files yet before we agree on something. Right now this PR:

- Removes all `ICSSProp`, `IHTMLProp`... interfaces, except for `IProp``
- Unifies `DOMAttr` and `HTMLProp`
- Changes the suffix `-Attr` to `-Prop`, so we now have only `IProp`, `SVGProp`, `CSSProp`, `DOMProp` and `FragmentProp`.

@forki @zaack @mangelmaxime @chadunit What do you think? Is this a good idea or just breaks code without much benefit? Any other suggestion?

### DIFF
--- a/src/Fable.React/Fable.Helpers.React.fs
+++ b/src/Fable.React/Fable.Helpers.React.fs
@@ -1,9 +1,7 @@
 module Fable.Helpers.React
 
-open System
 open System.Reflection
 open FSharp.Reflection
-open FSharp.Reflection.FSharpReflectionExtensions
 open Fable.Core
 open Fable.Core.JsInterop
 open Fable.Import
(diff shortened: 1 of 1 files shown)


zaaack:
I think SSR would  be OK if we only keep one  top interface, but unifies DOMAttr and HTMLAttr would cause the `renderHtmlAttr` and `renderSVGAttr`'s pattern match much longer if we want to keep pattern match's full case checking for better maintainance(which I'm hight recommend).

I have no advice about the naming， either attr or prop is OK for me.

MangelMaxime:
I think this is ok. But will need to check it over Fulma and one of my project to test the implication.

cc @et1975 too

chadunit:
I'm in favor of the unification since it cleans up additional explicit upcast scenarios.

I don't mind the original names, it might be good to keep them to reduce user code breakage (and just fold `DOMAttr` into `HTMLAttr`). I don't find mix of -Attr/-Prop too confusing since I feel they are generally accurate domain descriptions as used currently. I think `HTMLAttr` is especially likely to be frequently used in existing code, e.g. qualifying `HTMLAttr.Type` when it conflicts with `System.Type`. (Granted it's just a trivial rename so not a huge issue overall.)

alfonsogarciacaro:
Thanks a lot for your comments! I agree we shouldn't break code without a big reason. I'll close this for now, the big feature of Fable.React 3 is server rendering :)

PR #78, opened 2018-03-27
Adding DataTarget to HTML-Attributes

forki:
This attribute is needed for server-side rendering to render a menu button on mobile phones. Would appreciate a dot release. Thanks

/cc @horstoeko

### DIFF
--- a/src/Fable.React/Fable.Helpers.React.fs
+++ b/src/Fable.React/Fable.Helpers.React.fs
@@ -201,6 +201,7 @@ module rec Props =
         | Coords of string
         | CrossOrigin of string
         // | Data of string
+        | [<CompiledName("data-target")>] DataTarget of string
         | [<CompiledName("data-toggle")>] DataToggle of string
         | DateTime of string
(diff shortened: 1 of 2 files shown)


forki:
didn't know we can use the Data attribute

MangelMaxime:
@forki Is the `data-target` attribute in the html standard ? If yes, then we should add it to the binding.

If this is a custom `data-*`attribute then indeed we should use the `Data` attribute.

forki:
dunno

forki:
but it's used by bootstrap for mobile stuff

MangelMaxime:
Yes but bootstrap is evil :trollface: 

I did some search and in fact, we should not have `DataToggle` either the binding.

The standard say that we can create custom attribute if we need under `data-*`. 
[Specification](https://dev.w3.org/html5/spec-LC/elements.html#embedding-custom-non-visible-data-with-the-data-attributes)

Using `Data` helper is the right solution here.

PR #80, opened 2018-03-27
Update ssr benchmark

zaaack:


forki:
doesn't matter. good news is: we made ours much faster

Issue #84, opened 2018-04-15
Document process of using a third party react component


danyx23:
I think one of the big advantages of using Fable and React over e.g. Elm is that it is possible to use an ocean full of ready made React components. However, it is somewhat unclear what needs to be done to declare the required props etc. to use a third party React component. It would be great if this repo in the readme or the Fable docs could have a section on how to do this.

I don't know what the potential issues are with various kinds of e.g. Higher Order Components etc but it might be worth spelling these out explicitly as well. Please let me know if there is anything I can help with in particular to bring this forward.

alfonsogarciacaro:
Hi @danyx23! You're right, we have the pieces to easily import React components but we're missing the guide to do it. There are currently two ways to do it:

- **Dynamically**: Just import the component and define the properties with a pojo. A mini-guide to do that [can be found here](https://github.com/fable-compiler/fable-react/issues/61#issuecomment-368347998).
- **Creating bindings**: If you plan to use the component a lot, it maybe worthy to write some bindings to use DUs for the props and get more idiomatic F#. We're missing a guide for it (maybe @Zaid-Ajaj was writing one) although you can check other bindings in this repo ([like the one for Leaflet](https://github.com/fable-compiler/fable-react/blob/master/src/Fable.ReactLeaflet/ReactLeaflet.fs)). The most important thing is to call `keyValueList` before passing the prop list to the imported component ([as in here](https://github.com/fable-compiler/fable-react/blob/b236aa7c7cbdc18fac2b59d089878d355fd27e21/src/Fable.ReactLeaflet/ReactLeaflet.fs#L718-L719)).

danyx23:
Hi @alfonsogarciacaro, thank your for your reply! How do you suggest to go forward with this? Should I create a PR with an initial draft of a documentation document in the docs folder of this repo that outlines how to use a third party react component?

I am relatively new to Fable and have only used React a little bit - are there areas that are more tricky, like HOCs or similar somewhat advanced features of React?

alfonsogarciacaro:
This is funny, I was about to write that I hadn't used React HOCs so I didn't know, but just today I had to use them in my project and realized we're missing a helper in fable-react. I think we could add something like this:

```fsharp
open Fable.Core
open Fable.Import.React
open Fable.Helpers.React
open Fable.Helpers.React.Props

// Helper to be added to fable-react
[...]

PR #85, opened 2018-04-21
Document use of third party react components (#84)

danyx23:
This is a first stab at documenting how to use a third party react component. Please let me know what you think!

### DIFF
--- a/docs/using-third-party-react-components.md
+++ b/docs/using-third-party-react-components.md
@@ -1,8 +1,25 @@
-# Using third party React components
+## Using third party React components
 
-Using a third party (Javascript) React component is straightforward for most components. There are two ways of declaring a third party React component in F# - either by declaring a Pojo record for the props or by declaring a Discriminated Union where each case has one field.
+Using a third party (Javascript) React component is straightforward for most components. There are three ways of declaring a third party React component in F# - either by declaring a Discriminated Union where each case has one field; by declaring a record type for the props with the Pojo attribute; or by using an untyped list of `(string * obj)` tuples. All three ways are described below.
 
 Some components have a [Typescript](https://www.typescriptlang.org/) definition available, either because the component was authored in Typescript or someone created a type definition for the [Definitely Typed project](https://definitelytyped.org/). If this is the case then you can try the [ts2fable tool](https://github.com/fable-compiler/ts2fable) to convert this React component type definition from Typescript to a Fable type declaration - it might need some tweaking but for components with a big API surface this can be a real time saver.
+
+## Table of contents
+
+<!-- TOC -->
+
+- [Using third party React components](#using-third-party-react-components)
+- [Table of contents](#table-of-contents)
+- [Using a React component by declaring a Discriminated Union props type](#using-a-react-component-by-declaring-a-discriminated-union-props-type)
+  - [1. Install the react component](#1-install-the-react-component)
+  - [2. Define the props type](#2-define-the-props-type)
+  - [3. Define the React component creation function](#3-define-the-react-component-creation-function)
+  - [4. Use the creation function in your view code](#4-use-the-creation-function-in-your-view-code)
+- [Importing using a Pojo (plain old JS object) record](#importing-using-a-pojo-plain-old-js-object-record)
+- [Passing in props as tuples (without a type declaration of the props)](#passing-in-props-as-tuples-without-a-type-declaration-of-the-props)
+- [Edgecases](#edgecases)
+
+<!-- /TOC -->
 
 ## Using a React component by declaring a Discriminated Union props type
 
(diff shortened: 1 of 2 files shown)


docs/using-third-party-react-components.md
```diff
@@ -0,0 +1,90 @@
+# Using third party React components
+
+Using a third party (Javascript) React component is straightforward for most components. There are two ways of declaring a third party React component in F# - either by declaring a Pojo record for the props or by declaring a Discriminated Union where each case has one field.
+
+Some components have a [Typescript](https://www.typescriptlang.org/) definition available, either because the component was authored in Typescript or someone created a type definition for the [Definitely Typed project](https://definitelytyped.org/). If this is the case then you can try the [ts2fable tool](https://github.com/fable-compiler/ts2fable) to convert this React component type definition from Typescript to a Fable type declaration - it might need some tweaking but for components with a big API surface this can be a real time safer.
```
alfonsogarciacaro: Typo: time safer -> saver.

docs/using-third-party-react-components.md
```diff
@@ -0,0 +1,90 @@
...
+    strokeColor : string
+  }
+
+let inline progressLine (props : ProgressProps) (elems : ReactElement list) : ReactElement =
+    ofImport "Line" "rc-progress" props elems
+```
+
+## Edgecases
+
+This documentation needs to be extended to cover [Higher Order Components](https://reactjs.org/docs/higher-order-components.html) and maybe [Context]()https://reactjs.org/docs/context.html, [Fragments](https://reactjs.org/docs/fragments.html) etc. Contributions are welcome!
```
alfonsogarciacaro: Fix parens for the "Context" link.

docs/using-third-party-react-components.md
```diff
@@ -0,0 +1,90 @@
...
+For example to expose the percent, strokeWidth and strokeColor props of the rc-progress components:
+
+```fsharp
+type ProgressProps =
+  | Percent of int
+  | StrokeWidth of int
+  | StrokeColor of string
+```
+
+If one of the props is treated as a string enum in Javascript (e.g. if there is a size prop with the supported values "small", "normal" and "big"), then the `[<StringEnum>]` attribute can be very useful for defining helper types:
```
alfonsogarciacaro: Maybe adding a link to StringEnum in Fable docs? http://fable.io/docs/interacting.html#stringenum-attribute

docs/using-third-party-react-components.md
```diff
@@ -0,0 +1,90 @@
...
+```fsharp
+let view (model : Model) (dispatch : Msg -> unit) =
+  div
+    []
+    [ progressLine [ percent model.currentProgress; strokeColor: "red" ] [] ]
+```
+
+## Dynamic import using a Pojo
+
+The dynamic import is similar to the approach above, but instead of declaring a DU you create a Pojo record. This looks more like normal F# code but can be unwieldy if you have a lot of optional props (which is often the case in complex React components)
```
alfonsogarciacaro: Using the Pojo still makes the props typed, so I'd just say "Importing using a Pojo (plain old JS object) record" and then after this say you can declare props dynamically too:

```fsharp
open Fable.Core.JsInterop

ofImport "Line" "rc-progress" (createObj ["strokeWidth" ==> 5]) []
```

We could also add a note saying the Pojo attribute (link to Fable docs: http://fable.io/docs/interacting.html#plain-old-javascript-objects) is necessary because React only accepts plain JS objects (btw, the attribute won't be necessary in Fable 2, because records will compile directly to pojos).

docs/using-third-party-react-components.md
```diff
@@ -0,0 +1,90 @@
...
+  | Big
+
+type SomeComponentProps =
+  | Size of Size
+  | ...
+```
+
+### 3. Define the React component creation function
+
+Using the `ofImport` function you instruct Fable which component should be instantiated when the creation function is called.
```
alfonsogarciacaro: Maybe quickly warning about importing default members, which is a common source of confusion. You can also link to Fable docs: http://fable.io/docs/interacting.html#importing-javascript-code

alfonsogarciacaro:
This is great @danyx23, thanks a lot for this! :clap: :clap: :clap: I think your writing is very good, I just added a few minor comments. Please feel free to edit the text if you think they're helpful and then we merge the PR :)

danyx23:
@alfonsogarciacaro thanks, that was all good feedback. I tried to incorporate it all. Thanks to your hints I now understood the props/dynamic import much better and decided to break this out into it's own section; so now I write that there are three distinct ways of using a third parts react component.

I also added a table of contents because the document is already quite long now. Let me know if you think this is overkill and we should take it out again.

Please let me know if there is anything else you would like to see changed, otherwise I'm looking forward to the merge :)

docs/using-third-party-react-components.md
```diff
@@ -0,0 +1,126 @@
...
+The basic steps when working with a Discriminated Union are:
+
+### 1. Install the react component
+
+Using yarn or npm, install the react component you want to use.
+
+For example to use the [rc-progress React components](https://github.com/react-component/progress), run the following in your Fable project root:
+
+```bash
+yarn install rc-progress
```
inosik: This should be `yarn add`.

danyx23: Good catch, thanks

alfonsogarciacaro:
Awesome, thanks again for this!

PR #91, opened 2018-07-07
Fix namespace Fable.React => Fable.Helpers.React

0x53A:
I was really confused for an embarrassingly long time ;D


MangelMaxime:
Good catch thank you

PR #92, opened 2018-07-08
Put more emphasis on there being multiple exports/imports, add example for default import

0x53A:
I was really frustrated why I just imported ``undefined`` until I compared the generated Javascript against the example and noticed the extra ``{`` ``}``.

### DIFF
--- a/docs/using-third-party-react-components.md
+++ b/docs/using-third-party-react-components.md
@@ -99,7 +99,7 @@ you would declare your function like
 
 ```fsharp
 let inline qr_code_scanner (props : QRCodeScannerProps list) : ReactElement =
-    createElement(importDefault "react-native-qrcode-scanner", props, [])
+    createElement(importDefault "react-native-qrcode-scanner", (keyValueList CaseRules.LowerFirst props), [])
 ```
 
 ### 4. Use the creation function in your view code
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Thanks @0x53A! Yes, you're right. This is a problem that has hit many users (myself included). I still don't know if there's a way to type check this somehow (it's very unfortunate that Webpack doesn't display a good error message in these cases).

Instead of using `createElement(importDefault "foo", [], [])` it's also possible to just do `ofImport "default" "foo" [] []`.  Maybe we should recommend that instead? Actually I tried to do that in [latest Fable docs](https://github.com/fable-compiler/Fable/blob/2cd894d09a796a0ca805fe9326b9d7c28a210242/docs/interacting.md#importing-javascript-code) but I just noticed I forgot to update the website 🤦‍♂️ If you could update the PR @0x53A that'd be great, but I can also do it if you don't have time.

0x53A:
I didn't know about ``ofImport "default"``, thats why I didn't use it. :)

What do you think about the added third paragraph? Showing different equal ways is normally _bad_, but I think it may be helpful to show how to compose import and element instantiation.

alfonsogarciacaro:
Thanks @0x53A. I think the new paragraph is good. I will just add a sentence to make it clear Fable will group imports anyway so it's not a problem to duplicate `ofImport`, and probably also a reference to this: https://github.com/fable-compiler/fable-react/issues/60#issuecomment-368182485

PR #99, opened 2018-09-21
Initial version of Fable.ReactGoogleMaps

forki:
Hey @alfonsogarciacaro @MangelMaxime @isaaacabraham,

I'd like to finally OSS the react google maps, but I'd need some help.


src/Fable.ReactGoogleMaps/js/mapComponent.js
```diff
@@ -0,0 +1,91 @@
+import { withScriptjs, withGoogleMap, GoogleMap, TrafficLayer, Marker } from "react-google-maps";
+const { SearchBox } = require("react-google-maps/lib/components/places/SearchBox");
+import React, {Component, Children} from 'react';
+
+const TrafficMapComponent = withScriptjs(withGoogleMap((props) =>
+  <GoogleMap
```
forki: Is there a way to get rid of JSX? Can we rewrite this?

src/Fable.ReactGoogleMaps/Fable.Helpers.ReactGoogleMaps.fs
```diff
@@ -0,0 +1,172 @@
...
+
+/// A wrapper around google.maps.Map
+let googleMap (props:Props.IMapProperties list) : React.ReactElement =
+
+    R.from GoogleMapComponent (keyValueList CaseRules.LowerFirst props) []
+
+
+//
+// export function getPosition (options) {
+//     return new Promise(function (resolve, reject) {
```
forki: what's the easiest way to write that function directly in fable?

src/Fable.ReactGoogleMaps/Fable.Helpers.ReactGoogleMaps.fs
```diff
@@ -0,0 +1,172 @@
...
+    | Center of Coordinates.Position
+    | OnCenterChanged of (unit -> unit)
+    | OnPlacesChanged of (Places.Place [] -> unit)
+    | OnZoomChanged of (unit -> unit)
+    | OnIdle of (unit -> unit)
+    | Markers of React.ReactElement
+    | MapLoadingContainer of string
+    | MapContainer of string
+        interface IMapProperties
+let InfoWindow: RCom = import "InfoWindow" "react-google-maps"
```
forki: can we use fable to add this dependency automatically? if not I guess it's fine I write it in readme

src/Fable.ReactGoogleMaps/Fable.Helpers.ReactGoogleMaps.fs
```diff
@@ -0,0 +1,172 @@
...
+let marker (props:Props.IMarkerProperties list) children : React.ReactElement =
+    R.from Marker (keyValueList CaseRules.LowerFirst props) children
+
+let MarkerClusterer: RCom = import "MarkerClusterer" "react-google-maps/lib/components/addons/MarkerClusterer.js"
+
+/// A wrapper around MarkerClusterer - https://github.com/mahnunchik/markerclustererplus/blob/master/docs/reference.html
+let markerClusterer (props:Props.IMarkerClustererProperties list) (markers:React.ReactElement list) : React.ReactElement =
+    R.from MarkerClusterer (keyValueList CaseRules.LowerFirst props) markers
+
+let GoogleMapComponent: RCom = import "GoogleMapComponent" "./js/mapComponent.js"
```
forki: how can we make sure mapComponent.js is bundled? or can we rewrite as fable module?

src/Fable.ReactGoogleMaps/js/mapComponent.js
alfonsogarciacaro: Yes, it'll be better so users are not forced to include the Babel react-preset. It's really easy to replace JSX, you can use the [Babeljs repl](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=DwMQ9mAEBmELwCIBGBDATgyB6AfEA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=true&timeTravel=false&sourceType=module&lineWrap=false&presets=react%2Cstage-2%2Cstage-3%2Ces2015-loose%2Cenv&prettier=true&targets=Node-8.9&version=6.26.0&envVersion=1.6.2) to see the corresponding JS, but basically `<Foo foo="foo" bar={5}><Child1 /><Child2 /></Foo>` becomes:

```js
React.createElement(
  Foo,
  { foo: "foo", bar: 5 },
  React.createElement(Child1, null),
  React.createElement(Child2, null)
);
```

src/Fable.ReactGoogleMaps/Fable.Helpers.ReactGoogleMaps.fs
alfonsogarciacaro: You can use [Promise.create](https://github.com/fable-compiler/fable-powerpack/blob/1d107dc8312b38812947df975b2172f1de0ef2e1/src/Promise.fs#L17) from Fable.PowerPack.

src/Fable.ReactGoogleMaps/Fable.Helpers.ReactGoogleMaps.fs
alfonsogarciacaro: No, atm there's no mechanism to enforce npm dependencies. I was hoping Paket would do that for us ;)

src/Fable.ReactGoogleMaps/Fable.Helpers.ReactGoogleMaps.fs
alfonsogarciacaro: Just add it to the package in the `fable` folder keeping the same directory structure and it will work ([example](https://github.com/MangelMaxime/Thoth/blob/5bbe97540218d992289ca673b9593babafcc23bd/src/Thoth.Json/Thoth.Json.fsproj#L21)).

src/Fable.ReactGoogleMaps/Fable.Helpers.ReactGoogleMaps.fs
```diff
@@ -0,0 +1,172 @@
+module Fable.Helpers.ReactGoogleMaps
+
+open Fable.Core
+open Fable.Import
+open Fable.Core.JsInterop
+
+module Coordinates =
+
+    type [<Pojo>] Position = {
```
alfonsogarciacaro: Fable 2 doesn't do anything with the Pojo attribute any more. Seems React is not complaining now if you pass a non-plain JS object.

src/Fable.ReactGoogleMaps/Fable.Helpers.ReactGoogleMaps.fs
```diff
@@ -0,0 +1,172 @@
...
+    | OnIdle of (unit -> unit)
+    | Markers of React.ReactElement
+    | MapLoadingContainer of string
+    | MapContainer of string
+        interface IMapProperties
+let InfoWindow: RCom = import "InfoWindow" "react-google-maps"
+
+/// A wrapper around google.maps.InfoWindow
+let infoWindow (props:Props.IInfoWindowProperties list) child : React.ReactElement =
+    R.from InfoWindow (keyValueList CaseRules.LowerFirst props) [child]
```
alfonsogarciacaro: `keyValueList` accepts `seq` now. If you use `props: _ seq` instead of list in these signatures, Fable will have more chances to optimize the lists into arrays (or users can pass directly arrays if needed). Also if you inline these helpers, Fable will have more chances to resolve `keyValueList` at compile time instead of runtime.

```fsharp
let inline infoWindow (props:Props.IInfoWindowProperties seq) child : React.ReactElement =
    R.from InfoWindow (keyValueList CaseRules.LowerFirst props) [child]
```

src/Fable.ReactGoogleMaps/js/mapComponent.js
alfonsogarciacaro: Another option is to just pass this file through babel-cli to remove the JSX before packaging. I need to check but something like the following should work ([docs](https://babeljs.io/docs/en/babel-cli)). But remember to adjust the import path in the F# file:

```
npm install --save-dev @babel/core @babel/cli @babel/preset-react
npx babel script.js --out-file script-compiled.js --presets=react
```

forki: that would be awesome solution!

forki: ok got rid of the JSX. thanks!

forki:
how can I release it?

forki:
or @alfonsogarciacaro can you release it?

PR #101, opened 2018-09-25
Added FitBounds

nojaf:


src/Fable.ReactGoogleMaps/Fable.Helpers.ReactGoogleMaps.fs
```diff
@@ -58,6 +68,9 @@ type MapRef(mapRef) =
 
     member __.GetCenter() : Coordinates.Position =
         mapRef?getCenter() |> unbox
+        
+    member __.FitBounds(bounds: Coordinates.LatLngBounds) : unit =
```
forki: Can you please add an overload that takes the bounds that I already added in coordinates module?

src/Fable.ReactGoogleMaps/README.md
```diff
@@ -134,3 +134,33 @@ let myMap =
...
+
+```fs
+    // ...
+    let markerPositions: Position list = // ...
+        
+    // In some scenarios ref can be null, f.ex. slow network when the google maps isn't fully loaded yet.
+    let setRef (ref:obj) =
+        let bounds = ReactGoogleMaps.Coordinates.newLatLngBounds()
+        
+        match Option.ofObj ref with
```
forki: Just if isNull ref then would be simpler here

nojaf: Didn't know that one, but maybe the definition of `SetRef` should be `| SetRef of (obj option -> unit)`. That way it is explicit.

src/Fable.ReactGoogleMaps/README.md
```diff
@@ -134,3 +134,33 @@ let myMap =
...
+        | Some ref ->
+            markerPositions
+            |> List.fold (fun (acc:LatLngBounds) pos ->
+                acc.extend(!^ pos)
```
forki: what'S this doing?

src/Fable.ReactGoogleMaps/README.md
forki: well let me move the nullCheck into MapRef itself

src/Fable.ReactGoogleMaps/README.md
nojaf: Adds all `LatLngLiteral` to the newly created bounds. [extend](https://developers.google.com/maps/documentation/javascript/reference/coordinates#LatLngBounds.extend) returns `LatLngBounds`.

If you have say four markers on your map, you can add the locations to the bounds in order to know how to have the correct zoom and location so that your maps shows all four markers.

This is handy if you don't really know anything of your marker locations upfront.

PR #107, opened 2018-10-15
Use TextWriter for rendering to use as template engine

thinkbeforecoding:
I extracted the `render` function to provide it as a public `writeTo` that writes to a TextWriter.

The renterToString is now using a `StringWriter` which uses a `StringBuilder` internally, but for templating purpose, it is possible to use writeTo directly on a `StreamWriter` to write directly to files or network streams.


src/Fable.React/Fable.Helpers.ReactServer.fs
```diff
@@ -14,39 +14,36 @@ let private unitlessCssProps = System.Collections.Generic.HashSet<_>([ "animatio
...
+let private cssProp (html:TextWriter) (key: string) (value: obj) =
+  html.Write key
+  html.Write ':'
 
   match value with
   | :? int as v -> addUnit html key (string v)
   | :? float as v -> addUnit html key (string v)
-  | _ -> escapeHtml html (value.ToString()) |> ignore
-
-  html.Append ';' |> ignore
```
thinkbeforecoding: The cssProp function is not responsible for this anymore. It has been moved to renderHtmlAttr

src/Fable.React/Fable.Helpers.ReactServer.fs
```diff
@@ -614,20 +611,20 @@ let private renderHtmlAttr (html:StringBuilder) (attr: HTMLAttr) =
...
 
-    for cssProp in cssList do
-      renderCssProp html cssProp
-    if not(List.isEmpty cssList) then
-      html.Remove (html.Length - 1, 1) |> ignore
-    html.Append '"' |> ignore
+    cssList
+    |> List.iteri( fun i cssProp ->
+        if i > 0 then html.Write(';')
+        renderCssProp html cssProp)
```
thinkbeforecoding: Instead on removing the lase `;`, we write it before each prop except the first one.
TextWriter doesn't provide `Remove` and not writing in a first place is better anyway

thinkbeforecoding:
The goal is then to use this as a F# template engine for FSharp.Formatting for instance.

alfonsogarciacaro:
This looks like a great a addition and should expand the possibilities of SSR as you said. As the API is maintained I don't see any problem to merge it, but if @zaaack, as the SSR part author, and @forki, who has been using it, can have a look that would very helpful.

zaaack:
@thinkbeforecoding Great job! 

Just pulled and tried with the SSRSample app, updated the fable version.
here is the pr: https://github.com/thinkbeforecoding/fable-react/pull/1

forki:
Does it keep its perf characteristics?

forki:
@thinkb4coding if I understand correctly this can be used to convert a fable react site into a mostly static site, right? Here is a thought: our conference site which is built from https://github.com/fable-compiler/fableconf lacks good SEO. I always thought about putting it in SSR mode, but we don't have a proper Webserver set up for it. It's just github pages. If you are interested it would be awesome if you try your new tool on that project

thinkbeforecoding:
I use this tecnic for my blog:

It's hosted on azure functions but there are actually no functions. Just proxy rules to forward blobs that have been staticaly generated using Fsharp.Formatting for content and fable-react renderToString for templating.

So yes, doing it for this site is doable

thinkbeforecoding:
I merged the PR from @zaaack 

would there be a better way to mark the difference between renderToString that adds the reactroot attribute and the writeTo that doesn't ?


It could be possible to put the writeTo in a sub module to show it's more lowlevel stuff but could not find a good name.. so I let it there...

alfonsogarciacaro:
Thank you all! I merged the PR, but you're right @thinkbeforecoding, it may be better to add more semantics to the difference between `renderToString` and `writeTo`, I'll think of something.

@forki Actually the fable.io site is generated statically (unlike fableconf web which was created before) but I had to use a node server, this will allow faster iterations and using familiar IO commands from .NET. Thank you again @thinkbeforecoding!

thinkbeforecoding:
When do you think the nuget will be released ?

alfonsogarciacaro:
> When do you think the nuget will be released ?

I love your polite way to say: "Publish a nuget release NOW!!" :wink: Ok, I pushed Fable.React 4.1.1. As you suggested I put `writeTo` in a submodule named `Raw` to enforce the semantics and added a comment to make it clear the `reactroot` attribute is not added in that case :+1:

thinkbeforecoding:
Perfect ! 😄 Thx !!

PR #109, opened 2018-10-23
Make castHTMLNode public for use with Raw.WriteTo

thinkbeforecoding:
Raw.writeTo accepts a HTMLNode while constructed model outputs a ReactElement.

The code already provides a private castHTMLNode function for conversion.

This PR make it public to be used from the outside.

The implementation has also been changed to make it safer. Any non matching input now returns a HTMLNode.Empty instead of throwing an exception.


alfonsogarciacaro:
Thanks!

PR #112, opened 2018-11-05
Add React.memo

### LINKED ISSUE #111: React.memo
It would be nice if we had a helper for [React.memo](https://reactjs.org/docs/react-api.html#reactmemo).
I believe @vbfox had already suggested a possible approach at FableConf. Any news on this?

vbfox:
It also add to the helper types the concept of a `ReactComponentType` (Something that can be passed to createElement). Some of the types exists in the typescript-imported part of the library but they aren't really adapted and try to handle lot of things that are now obsolete.

They will also be necessary to import `React.lazy`

I didn't add any higher level helpers but one I find useful is the capability to create a named memo without passing the comparer :

```fsharp
let namedMemo<'props> (name: string) (f: 'props -> ReactElement): ReactComponentType<'props> =
    f?displayName <- name
    memo(f, None)

// Usage
type HelloProps = { Name: string }
let helloWorldComponent = namedMemo "HelloWorld" (fun props ->
  span [] [str "Hello "; str props.Name]
)
let helloWorld name = ofElementType helloWorldComponent { Name = name } []
```

Should I add it too ?

PS: *No server-side rendering for now as I don't know how it should work.*

Fixes #111

### DIFF
--- a/src/Fable.React/Fable.Helpers.React.fs
+++ b/src/Fable.React/Fable.Helpers.React.fs
@@ -851,6 +851,14 @@ let inline fn<'P> (f: 'P -> ReactElement) (props: 'P) (children: ReactElement se
 let inline ofImport<'P> (importMember: string) (importPath: string) (props: 'P) (children: ReactElement seq): ReactElement =
     createElement(import importMember importPath, props, children)
 
+[<Erase>]
+type ReactElementType<'props> = interface end
+
+[<Erase>]
+type ReactComponentType<'props> =
+    inherit ReactElementType<'props>
+    abstract displayName: string option with get, set
+
 #if FABLE_COMPILER
 /// Alias of `ofString`
 let inline str (s: string): ReactElement = unbox s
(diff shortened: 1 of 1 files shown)


src/Fable.React/Fable.Helpers.React.fs
```diff
@@ -851,6 +851,14 @@ let inline fn<'P> (f: 'P -> ReactElement) (props: 'P) (children: ReactElement se
 let inline ofImport<'P> (importMember: string) (importPath: string) (props: 'P) (children: ReactElement seq): ReactElement =
     createElement(import importMember importPath, props, children)
 
+[<Erase>]
+type ReactElementType<'props> = interface end
+
+[<Erase>]
+type ReactComponentType<'props> =
+    inherit ReactElementType<'props>
+    abstract displayName: string option with get, set
```
alfonsogarciacaro: No need to add `Erase` attribute to interfaces. They're already erased from generated code.

src/Fable.React/Fable.Helpers.React.fs
```diff
@@ -878,6 +886,40 @@ let inline ofList (els: ReactElement list): ReactElement = unbox(List.toArray el
...
+/// classes.
+///
+/// If your function component renders the same result given the same props, you can wrap it in a call to React.memo
+/// for a performance boost in some cases by memoizing the result. This means that React will skip rendering the
+/// component, and reuse the last rendered result.
+///
+/// By default it will only shallowly compare complex objects in the props object. If you want control over the
+/// comparison, you can also provide a custom comparison function as the second argument.
+[<Import("memo", from="react")>]
+let memo<'props> (render: 'props -> ReactElement, areEqual: PropsEqualityComparison<'props> option) : ReactComponentType<'props> =
```
alfonsogarciacaro: Instead of forcing users to write `memo (renderFunction, None)` maybe we could offer two different functions:

```fsharp
let memo renderFn = import "memo" "react"
let memoWith renderFn areEqual = import "memo" "react"
```

MangelMaxime:
I think you should add the `namedMemo` function.

alfonsogarciacaro:
Thanks @vbfox! Looks good :+1: I just left a couple of comments. It would be nice to have `namedMemo` too, although I'm not sure how to mix that with my proposal of having two functions `memo` and `memoWith` 🤔

alfonsogarciacaro:
Another option could be to use a static type helper so we can use optional arguments (areEqual, name)... But I know @MangelMaxime doesn't like optional arguments XD

alfonsogarciacaro:
Or a list of attributes in line with other React components:

```fsharp
memo [Name "foo"; AreEqual (fun p1 p2 -> p2.x = p2.x)] (fun props ->
   ...)
```

vbfox:
```fsharp
type HelloProps = { Name: string }

let helloWorldComponent = memo "HelloWorld" (fun props ->
  span [] [str "Hello "; str props.Name]
)
let helloWorld name = ofElementType helloWorldComponent { Name = name } []

let helloWorldComponent2 = memoWith "HelloWorld" (fun a b -> a.Name = b.Name) (fun props ->
  span [] [str "Hello "; str props.Name]
)
let helloWorld2 name = ofElementType helloWorldComponent2 { Name = name } []
```

alfonsogarciacaro:
Awesome, thanks a lot!

alfonsogarciacaro:
I've published this as Fable.React 5.0.0-alpha-001 (at the end I used `memoBuilder` for the name, sorry!). I've also tried to adapt it to SSR, @forki can you please give it a try? Thanks!

PR #113, opened 2018-11-05
Add nothing to react helpers

vbfox:
React use `null` to signal that nothing should be rendered, it could be done before with `ofOption None` but this add clearer alternative to generate this.

```fsharp
let render props =
	let details =
		match props.Details with
		| Some d -> strong [] [str (sprintf " (%s)")] d
		| None -> nothing
	span [] [ str props.Name; details ]
```

### DIFF
--- a/src/Fable.React/Fable.Helpers.React.fs
+++ b/src/Fable.React/Fable.Helpers.React.fs
@@ -878,6 +878,10 @@ let inline ofList (els: ReactElement list): ReactElement = unbox(List.toArray el
 /// Returns an array **from .render() method**
 let inline ofArray (els: ReactElement array): ReactElement = unbox els
 
+/// A ReactElement when you don't want to render anything (null in javascript)
+[<Emit("null")>]
+let nothing: ReactElement = jsNative
+
 #else
 /// Alias of `ofString`
 let inline str (s: string): ReactElement = HTMLNode.Text s :> ReactElement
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Nice one, thank you!

forki:
awesome this reduces couple of yields for me. How can we make it mor popular? @theimowski something for the template?

theimowski:
Yeah, maybe - created an issue

MangelMaxime:
Just for reference you can also use `Option.map` to not use `yield`.

```fs
let private showFirstConnectionUrl (url : string) =
    Message.message [ Message.Color IsInfo ]
        [ Message.header [ ]
            [ str "Ce compte n'a pas encore été activé." ]
          Message.body [ ]
            [ CopyButton.copyButtton [ CopyButton.Value url ] ] ]

let private viewUserInfo (info : LoadedModel) dispatch =
    div [ ]
        [ info.OriginalData.FirstConnectionUrl
          |> Option.map showFirstConnectionUrl
          |> ofOption
          // More code... 
        ]
```

forki:
Yes, but really like the explicit nothing function.

Am Di., 6. Nov. 2018, 11:05 hat Maxime Mangel <notifications@github.com>
geschrieben:

vbfox:
Yes using `Option.Map + ofOption` is the other choice but for complex case I find `nothing` to be clearer and producing less noise instead having to add `Some`/`None` + `ofOption`, when starting with options in the model then it depends

PR #119, opened 2018-11-06
Simplify memo usage by default

vbfox:
This is a breaking change to memo and memoWith !

They become builder functions, producing functions that can be directly used in a render method and don't support children.

The high-level version are still available in the ReactElementType module for anyone wanting the more complete experience (Like children support, or later returning component types themselves to use in `lazy`).

Sample usage :
```fsharp
type HelloProps = { Name: string }
let hello = memo "Hello" <| fun { Name = name } ->
    span [str "Hello "; str name]

let view model =
    hello { Name = model.Name }
```

### DIFF
--- a/src/Fable.React/Fable.Helpers.React.fs
+++ b/src/Fable.React/Fable.Helpers.React.fs
@@ -915,7 +915,7 @@ module ReactElementType =
     /// By default it will only shallowly compare complex objects in the props object. If you want control over the
     /// comparison, you can use `memoWith`.
     [<Import("memo", from="react")>]
-    let inline memo<'props> (render: 'props -> ReactElement) : ReactComponentType<'props> =
+    let memo<'props> (render: 'props -> ReactElement) : ReactComponentType<'props> =
         jsNative
 
     [<Import("memo", from="react")>]
--- a/src/Fable.React/Fable.Helpers.React.fs
+++ b/src/Fable.React/Fable.Helpers.React.fs
@@ -888,6 +888,8 @@ let inline ofArray (els: ReactElement array): ReactElement = unbox els
 [<Emit("null")>]
 let nothing: ReactElement = jsNative
 
+type PropsEqualityComparison<'props> = 'props -> 'props -> bool
+
 [<RequireQualifiedAccess>]
 module ReactElementType =
     let inline ofComponent<'comp, 'props, 'state when 'comp :> Component<'props, 'state>> : ReactComponentType<'props> =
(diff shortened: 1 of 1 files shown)


src/Fable.React/Fable.Helpers.React.fs
```diff
@@ -903,45 +905,82 @@ module ReactElementType =
...
+/// let hello = memo "Hello" (fun { Name = name } -> span [str "Hello "; str name])
+///
+/// let view model =
+///     hello { Name = model.Name }
+/// ```
 ///
 /// By default it will only shallowly compare complex objects in the props object. If you want control over the
 /// comparison, you can use `memoWith`.
-let memo<'props> (name: string) (render: 'props -> ReactElement) : ReactComponentType<'props> =
+let memo<'props> (name: string) (render: 'props -> ReactElement) : 'props -> ReactElement =
```
alfonsogarciacaro: Should we use the name `memoBuilder` here? This make it more obvious that the function is not a direct binding of `memoBuilder`, then we keep the `memo` direct binding in `ReactElementType` module for people who need it.

src/Fable.React/Fable.Helpers.React.fs
```diff
@@ -903,45 +905,82 @@ module ReactElementType =
...
+    /// classes.
+    ///
+    /// If your function component renders the same result given the same props, you can wrap it in a call to React.memo
+    /// for a performance boost in some cases by memoizing the result. This means that React will skip rendering the
+    /// component, and reuse the last rendered result.
+    ///
+    /// This version allow you to control the comparison used instead of the default shallow one by provide a custom
+    /// comparison function.
+    let memoWith<'props> (areEqual: PropsEqualityComparison<'props>) (render: 'props -> ReactElement) : ReactComponentType<'props> =
+        reactMemoWith(render, areEqual)
```
alfonsogarciacaro: The F# compiler (and thus Fable) translates module functions with tuple and curried arguments in the same way, so you should be able to add the `Import` attribute directly here.

alfonsogarciacaro:
Thanks @vbfox, looks good and comments are great! I just left a couple of suggestions.

Fable.React 4.1.3 package has already been downloaded a few times so it may be better not to unlist it. I wonder if we should start the 5 cycle and release this as alpha. Then we could also begin adding helpers for the other new React APIs (lazy, hooks) or it may be too soon?

src/Fable.React/Fable.Helpers.React.fs
vbfox: Yes but I can't reverse the arguments :)

src/Fable.React/Fable.Helpers.React.fs
vbfox: I think the comment with sample is  enough to defuse the confusion

vbfox:
I don't think anyone who downloaded the package would have used memo... Beside me XD

Lazy I will add after fully testing it. Might be too soon for hooks it is still a proposal subject to change

src/Fable.React/Fable.Helpers.React.fs
alfonsogarciacaro: Ah, didn't notice that, sorry! 😅 Then this works :+1: There's still a way to use only attributes, but it's a bit hacky and I haven't used it very much:

```fsharp
[<Import("memo", from="react"); Emit("$0($2, $1)")>]
let memoWith<'props> (areEqual: PropsEqualityComparison<'props>) (render: 'props -> ReactElement) : ReactComponentType<'props> = jsNative
```

forki:
I want to. so just get the breaking stuff in so that I can start! :P

Am Mi., 7. Nov. 2018 um 11:36 Uhr schrieb Julien Roncaglia <
notifications@github.com>:

src/Fable.React/Fable.Helpers.React.fs
vbfox: I love Emit but it might be a little bit too hacky :)

PR #139, opened 2019-01-25
Add aria-controls and aria-pressed HtmlAttributes

rfrerebe:
See :
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role


alfonsogarciacaro:
Thank you!

rfrerebe:
Thanks for merging 😄

PR #144, opened 2019-02-23
Add Label property for Marker

markpattison:


alfonsogarciacaro:
I assume this is OK so I'll merge it. But I haven't used the React Google Maps library so I won't be able to review more complicated reviews. Maybe we should move this package to its own repo? @forki

PR #147, opened 2019-03-24
Add type-safe css properties

Zaid-Ajaj:
This PR adds a couple of type definitions to make usage of CSS properties type-safe. It is not much but it only covers a couple of commonly used inline styles:
 - [x] Display 
 - [x] Position
 - [x] TextAlign
 - [x] AlignContent
 - [x] AlignItems 
 - [x] AlignSelf 

This is of course a *breaking* change because the properties such as `Display` were taking `obj` as input where the user supplies a string a everything works out (hopefully) but now the types are restricted to the defined DU's

This PR doesn't affect bundle sizes. All DU's are simple `StringEnum`s. 

I thought it would be a good time to work on these breaking changes since we are already updating to a major version.

### DIFF
--- a/src/Fable.React.Props.fs
+++ b/src/Fable.React.Props.fs
@@ -318,10 +318,151 @@ type HTMLAttr =
 #endif
     interface IHTMLProp
 
+[<StringEnum; RequireQualifiedAccess>]
+/// Specifies the display behavior (the type of rendering box) of an element.
+type Display = 
+    /// The element is completely removed
+    | None 
+    /// Sets this property to its default value. Read more about [CSS Initial value](https://www.w3schools.com/CSSref/css_initial.asp)
+    | Initial
+    /// Inherits this property from its parent element. Read about [CSS Inherit](https://www.w3schools.com/CSSref/css_inherit.asp)
+    | Inherit
+    /// Displays an element as an inline element (like `<span>`). Any height and width properties will have no effect.
+    | Inline 
+    /// Displays an element as a block element (like `<p>`). It starts on a new line, and takes up the whole width.
+    | Block
+    /// Makes the container disappear, making the child elements children of the element the next level up in the DOM.
+    | Contents 
+    /// Displays an element as a block-level flex container.
+    | Flex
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
You're right, much better to use stringenums, thank you!

alfonsogarciacaro:
I'm getting some conflict with the `Display` StringEnum and [Fulma's `Display.Option`](https://github.com/Fulma/Fulma/blob/f1939b452f8295eda4c3a4775d611c68caf46db1/src/Fulma/Common.fs#L219). The weird thing is when I try to fully qualify the type like `Fable.React.Props.Display.Flex` the compiles thinks this is the CSSProp.

The only solution I've found to fix it was using a type alias: `type CSSDisplay = Fable.React.Props.Display`

Zaid-Ajaj:
Does it work if you give the compile a hint that you are using `CSSProp`? i.e.
```fs
div [ CSSProp.Display Display.Flex ] [ ]
```

> The weird thing is when I try to fully qualify the type like Fable.React.Props.Display.Flex the compiles thinks this is the CSSProp.

F# compiler or Fable? sounds like a bug to me because you are using a constructor of `Display` type (the `Flex` case) and not a constructor of `CSSProp` (the `Display` case)

alfonsogarciacaro:
> Does it work if you give the compile a hint that you are using CSSProp? i.e.

Nope, same issue :/

> F# compiler or Fable?

I can see the error both in the IDE and Fable's log output

MangelMaxime:
I saw this problem occurred several times in my projects.

I tried to re-order the `open` statement or to fully qualified the whole path but it didn't work.

alfonsogarciacaro:
I will add "-Options" to the StringEnums to avoid this issue (DisplayOptions, TextAlignOptions, etc). Does it look right?

Zaid-Ajaj:
SGTM, but still this looks like a bug in type-inference for cross-project types. *We* need to report it somewhere, maybe f# compiler repo or track in Fable repo

PR #152, opened 2019-04-17
Name memo elements

vbfox:
Memo when used with a comparison function create a named node
in the React debugger.

The node is named 'Memo' by default and we change it to
'Memo(displayName)' if a displayName is passed to FunctionComponent.Of

image


alfonsogarciacaro:
Good point, thanks you!

PR #153, opened 2019-04-17
Faster equality

vbfox:
* Change `equalsButFunctions` to stop early when a difference is found instead of continuing to compare the rest of the properties
* Provide another variant `memoEqualsButFunctions` using reference equality instead of F# one. There result is that it should do the same thing as the the default memo equality except that it ignore functions

### DIFF
--- a/src/Fable.React.Helpers.fs
+++ b/src/Fable.React.Helpers.fs
@@ -204,6 +204,31 @@ module Helpers =
         false
 #endif
 
+    /// Same as the default `React.memo` equality but ignores functions.
+    /// Useful in combination with `FunctionComponent.Of` `memoizeWith` parameter when F# equality isn't required
+    let memoEqualsButFunctions (x: 'a) (y: 'a) =
+#if FABLE_COMPILER
+        if obj.ReferenceEquals(x, y) then
+            true
+        elif isNonEnumerableObject x && not(isNull(box y)) then
+            let keys = JS.Object.keys x
+            let length = jsArrayLength(keys)
+            let mutable i = 0
+            let mutable result = true
+            while i < length && result do
+                let key = keys.[i]
+                i <- i + 1
+                let xValue = x?(key)
+                result <- isFunction xValue || obj.ReferenceEquals(xValue, y?(key))
+            result
+        else
+            obj.ReferenceEquals(box x, box y)
+#else
+        // Server rendering, won't be actually used
+        // Avoid `x = y` because it will force 'a to implement structural equality
+        false
+#endif
+
     [<System.Obsolete("Use FunctionComponent.Of with memoizedWith")>]
     let memoBuilder<'props> (name: string) (render: 'props -> ReactElement) : 'props -> ReactElement =
 #if FABLE_COMPILER
--- a/src/Fable.React.Helpers.fs
+++ b/src/Fable.React.Helpers.fs
@@ -173,18 +173,31 @@ module Helpers =
 
     [<Emit("typeof $0 === 'object' && !$0[Symbol.iterator]")>]
     let private isNonEnumerableObject (x: obj): bool = jsNative
+
+    [<Emit("$0.length")>]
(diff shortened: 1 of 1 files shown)


src/Fable.React.Helpers.fs
```diff
@@ -173,18 +173,56 @@ module Helpers =
 
     [<Emit("typeof $0 === 'object' && !$0[Symbol.iterator]")>]
     let private isNonEnumerableObject (x: obj): bool = jsNative
+
+    [<Emit("$0.length")>]
+    let private jsArrayLength(arr: ResizeArray<_>): int = jsNative
```
alfonsogarciacaro: Hmm, `.Count` has some overhead (it has to checker whether the object is actually an array, although maybe we could optimize that), but if we're using so many JS helpers I wonder if it wouldn't be better to write the function directly in JS.

src/Fable.React.Helpers.fs
```diff
@@ -173,18 +173,56 @@ module Helpers =
...
+            (box x) = (box y)
+#else
+        // Server rendering, won't be actually used
+        // Avoid `x = y` because it will force 'a to implement structural equality
+        false
+#endif
+
+    /// Same as the default `React.memo` equality but ignores functions.
+    /// Useful in combination with `FunctionComponent.Of` `memoizeWith` parameter when F# equality isn't required
+    let memoEqualsButFunctions (x: 'a) (y: 'a) =
```
alfonsogarciacaro: If you don't mind, we could change the name to `refEqualsButFunctions` to make the semantics clearer. I'm afraid it may not be immediately obvious to users this is closer to how React.memo default equality works (and they may not know what that means).

src/Fable.React.Helpers.fs
vbfox: Yeah I removed Count when I saw the codegen... as for writting it in JS directly it make the calls to equal a little more complex but yeah it might be easier.

I just wasn't sure how to add pure-JS code to a pure-F# library. (Especially handling server-side rendering where there should still be an empty version for the .NET F# compiler to find

src/Fable.React.Helpers.fs
vbfox: That was my first name but I feared that reference equality would be misleading.

The name doesn't exists in JS so JS/React developers wouldn't guess it and as it uses `===` it's actually not doing a reference equality check, strings in particular are compared by value (something we very much want)

Memo isn't perfect because it assumes JS/React knowledge either but it sounded less of a subtle lie.

Anyway feel free to change, I don't think there is any perfect answer there it's on the fence between JS and .NET communities :)

alfonsogarciacaro: Hmm, you're right. Actually you made me think because in Fable we translate `obj.ReferenceEquals` as `===` but this gives different results in .NET and Fable for value types and strings:

```fsharp
let x = 5
let y = 5

// false in .NET, true in Fable
obj.ReferenceEquals(x, y) |> printfn "%b"
```

Should we use another name, `primitiveEquals` or similar?

vbfox: I think it's ok that obj.ReferenceEquals do is as AFAIK true reference equality isn't avaiable in JS so it's the nearest thing.

`primitiveEquals` seem good as it doesn't have the mental association with .NET reference equality

alfonsogarciacaro: It seems the closes function in F# to JS `===` behavior is [LanguagePrimitives.PhysicalEquality](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/languageprimitives.physicalequality%5B't%5D-function-%5Bfsharp%5D). Should we use `physicalEquality`?

I doubt it, but we also should be careful in case obj.ReferenceEquals implementation in Fable changes in the future. Maybe we should just write the function in JS at the end :/ And also probably you're right `memoEquals` may be the best name ;)

alfonsogarciacaro:
Release as 5.1.0 :)

PR #158, opened 2019-04-21
Clarify function docs

cmeeren:
I find these descriptions a bit more clear. But please check for errors; I may have misunderstood some details.

By the way, how do both of these functions work with tuples? Perhaps that should be mentioned, too?

### DIFF
--- a/src/Fable.React.Helpers.fs
+++ b/src/Fable.React.Helpers.fs
@@ -176,9 +176,9 @@ module Helpers =
     let private isNonEnumerableObject (x: obj): bool = jsNative
 #endif
 
-    /// Same as F# equality but ignores functions in the first level of an object.
-    /// Useful in combination with `FunctionComponent.Of` `memoizeWith` parameter for most cases
-    /// (ignore Elmish dispatch, etc)
+    /// Memberwise structural comparison (same as default F#) that ignores top-level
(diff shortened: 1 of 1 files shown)


cmeeren:
Sorry for the rogue EOL edit; the GitHub editor threw that in for free...

src/Fable.React.Helpers.fs
```diff
@@ -202,8 +202,9 @@ module Helpers =
         false
 #endif
 
-    /// Same as the default `React.memo` equality but ignores functions.
-    /// Useful in combination with `FunctionComponent.Of` `memoizeWith` parameter when F# equality isn't required
+    /// Memberwise referential comparison (same as default React.memo) that ignores
```
alfonsogarciacaro: I agree the comment is not very clear but the tricky part is this is not exactly referential comparison (for numbers and strings for example the comparison is by value). There's some [discussion about it here](https://github.com/fable-compiler/fable-react/pull/153#discussion_r276569792).

cmeeren: Better now? Note that I use the term "value types", but I'm not sure exactly how the equality works with structs (e.g. `DateTime`) in the JS environment. Are they compared by value?

alfonsogarciacaro:
Thanks!

PR #159, opened 2019-04-21
Clarify what FunctionComponent.Of does

cmeeren:
Reading through the [Fable.React 5 blog post](https://fable.io/blog/Announcing-Fable-React-5.html), I was confused by the fact that `React.memo` wasn't mentioned anywhere. Digging into the code, it seems to me that the new `FunctionComponent.Of` uses `React.memo` when `memoizeWith` is specified. I think it would be best to mention this is the function's documentation.


alfonsogarciacaro:
Yes, I missed the documentation comment for `FunctionComponent.Of` when publishing Fable.React 5. I should write something a bit more comprehensive, thank you!

Issue #164, opened 2019-06-03
How should I use the Context api?


0x53A:
I am currently working on a localization api for fable react.

What I want is to be able to set the language somewhere at the top, and then have the child nodes reflect this language. For that, the Context api seems to be a perfect fit.

So I want to do the equivalent to

```jsx
const LocalizationContext = React.createContext( { lang: "en" } );

// ...

// app
  render() {
    return (
      <Page>
        <LocalizationContext.Provider value={{ lang: "it" }}>

			/* imagine a few layers of sub-components and nesting */

			<LocalizationContext.Consumer>
				{value => /* render localized text based on value.lang */}
			</LocalizationContext.Consumer>

        </LocalizationContext.Provider>
      </Page>
    );
  }
  
```

Now, what initially confused me is that in Fable.React, the interface ``IContext`` is empty:

_(I would have expected two members, Provider and Consumer)_

https://github.com/fable-compiler/fable-react/blob/3dfbd761f1b01cba0934a6f2aef3ca918a14f6bc/src/Fable.React.fs#L18-L19

Then a bit later I found ``contextProvider``:

https://github.com/fable-compiler/fable-react/blob/3dfbd761f1b01cba0934a6f2aef3ca918a14f6bc/src/Fable.React.Helpers.fs#L345-L353

So that takes care of the Provider part.

--------

But I couldn't find anything like ``contextConsumer``.

How should I consume the context? Did I miss anything?

I'm not 100% sure how to translate

```jsx
			<LocalizationContext.Consumer>
				{value => /* render localized text based on value.lang */}
			</LocalizationContext.Consumer>
```

to Fable React, afaik the function should be passed as a prop named "children"? I think the following should work:

```fs
     let inline contextConsumer (ctx: IContext<'T>) (children: 'T -> ReactElement seq): ReactElement = 
#if FABLE_COMPILER 
         ReactBindings.React.createElement(ctx?Consumer, createObj ["children " ==> children ], null) 
#else 
         children((ctx :?> ISSRContext<_>).DefaultValue)
#endif 
```

If I manage to get anything working I'll send a PR ...

alfonsogarciacaro:
[...]

PR #165, opened 2019-06-03
add 'contextConsumer'

0x53A:
closes https://github.com/fable-compiler/fable-react/issues/164

~~Note: this is currently completely untested.~~

### DIFF
--- a/src/Fable.React.Helpers.fs
+++ b/src/Fable.React.Helpers.fs
@@ -343,6 +343,7 @@ module Helpers =
 
     /// Accepts a context value to be passed to consuming components that are descendants of this Provider.
     /// One Provider can be connected to many consumers. Providers can be nested to override values deeper within the tree.
+    /// Important: In SSR, this is ignored, and the DEFAULT value is consumed!
     /// More info at https://reactjs.org/docs/context.html#contextprovider
     let inline contextProvider (ctx: IContext<'T>) (value: 'T) (children: ReactElement seq): ReactElement =
 #if FABLE_COMPILER
@@ -351,7 +352,7 @@ module Helpers =
         fragment [] children
 #endif
 
-    /// Consumes a context value either from the nearest parent in the tree, or from the default value.
+    /// Consumes a context value, either from the nearest parent in the tree, or from the default value.
     /// Important: in SSR, this will always consume the context DEFAULT value!
     /// More info at https://reactjs.org/docs/context.html#contextconsumer
     let inline contextConsumer (ctx: IContext<'T>) (children: 'T -> ReactElement): ReactElement = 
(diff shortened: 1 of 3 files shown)


0x53A:
It currently works on the client side, but not in SSR:

```f#
let ctx = createContext("de")

let inline consume() = contextConsumer ctx (fun lang -> str lang)

// ...

let view model =

// ...

                str "This should be the default value (de):"
                consume()
                contextProvider ctx model.Language [
                    str (sprintf "This should be the model value (%s):" model.Language)
                    consume()
                    contextProvider ctx "xyz" [
                        str (sprintf "This should be the overridden value (%s):" "xyz")
                        consume()
                    ]
                    str (sprintf "This should be the model value (%s):" model.Language)
                    consume()
                ]
                str "This should be the default value (de):"
                consume()

```


Clientside:

image


SSR:

image



I think it renders the elements too early, so that my stack pushing/popping is completely useless ...

Since SSR is currently incorrect, anyway, this is not a regression, so i think I'll revert that part and just PR  the new 'contextConsumer' function.

What do you think?

alfonsogarciacaro:
This is due to my "cheap solution" in order to support the context API in .NET without many code changes. As it's currently difficult to pass a context object through the nested rendered elements in SSR, I'm just using the default value and _hoping_ the proper one will be used on the client side when mounting React. Does it happen like this in your example?

Maybe it'd be better to fix it in SSR (your note in the comment is already helpful!) but I'm not sure what would be the easiest way to do this. Any ideas?

In any case, thanks for the PR!

src/Fable.React.Helpers.fs
```diff
@@ -351,6 +352,17 @@ module Helpers =
         fragment [] children
 #endif
 
+    /// Consumes a context value, either from the nearest parent in the tree, or from the default value.
+    /// Important: in SSR, this will always consume the context DEFAULT value!
+    /// More info at https://reactjs.org/docs/context.html#contextconsumer
+    let inline contextConsumer (ctx: IContext<'T>) (children: 'T -> ReactElement): ReactElement = 
+#if FABLE_COMPILER 
+        ReactBindings.React.createElement(ctx?Consumer, createObj [], [!!children])
```
alfonsogarciacaro: I think we can pass `null` here instead of `createObj []`.

0x53A:
> 'm just using the default value and hoping the proper one will be used on the client side when mounting React. Does it happen like this in your example?

Yes, after the first client-side render all is ok.

In my use-case (localization) it doesn't matter anyway, there is only one context anyway.

> Maybe it'd be better to fix it in SSR (your note in the comment is already helpful!) but I'm not sure what would be the easiest way to do this. Any ideas?

No simple idea, the rendering would need to be delayed so that the parent can influence the child.

alfonsogarciacaro:
Ok, if the client side fixes this let's consider it good enough for now ;) Released as 5.2.4, thank you!

PR #168, opened 2019-06-28
Add forwardRef ?

Luiz-Monad:
It would be useful to be able to forwardRef when creating Functional Components.


alfonsogarciacaro:
Thanks!

PR #169, opened 2019-07-03
Npm packages metadata

Zaid-Ajaj:
Packages validated with Femto locally, using version constraint `>= 16.8.0` because that is where React shipped hooks, downstream package [Fable.Elmish.React](https://github.com/elmish/react) will need an update too along with [Fable.ReactNative](https://github.com/fable-compiler/fable-react-native) because they both depend on this package

@alfonsogarciacaro @forki  @MangelMaxime

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -258,3 +258,4 @@ packages/
 .idea/
 *.sln.iml
 *.orig
+.ionide
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,7 @@
+### 5.3.0
+
+- Add npm dependency metadata to work with Femto
+
 ### 5.2.6
 
 - Fix #167: Add `withKey` argument to `FunctionComponent.Of`
--- a/src/Fable.React.fsproj
+++ b/src/Fable.React.fsproj
@@ -7,6 +7,12 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <!-- <DefineConstants>$(DefineConstants);FABLE_COMPILER</DefineConstants> -->
   </PropertyGroup>
+  <PropertyGroup>
+    <NpmDependencies>
+        <NpmPackage Name="react" Version=">= 16.8.0" />
+        <NpmPackage Name="react-dom" Version=">= 16.8.0" />
+    </NpmDependencies>
+  </PropertyGroup>
   <ItemGroup>
     <Compile Include="Fable.React.fs" />
     <Compile Include="Fable.ReactDom.fs" />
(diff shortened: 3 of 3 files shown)


MangelMaxime:
Here we are hitting a problem.

`react-native` doesn't need `react-dom` in order to works.

See [package.json](https://github.com/SAFE-Stack/SAFE-Nightwatch/blob/master/package.json) from [SAFE-Nightwatch](https://github.com/SAFE-Stack/SAFE-Nightwatch/)

MangelMaxime:
For me [Fable.ReactDom.fs](https://github.com/fable-compiler/fable-react/blob/master/src/Fable.ReactDom.fs) should be in it's own package.

This changes is following the same philosophy applied to Fable.Browser.*, Fulma, etc.

Zaid-Ajaj:
I was thinking the exact same thing, the hierarchy should be 
```
Fable.Elmish.React
  |
  | -- Fable.ReactDom
           |
           | -- Fable.React
```
Same goes for `Fable.ReactNative`, but I am not sure if `Fable.Elmish.React` is specific to `Fable.ReactDom` or just `Fable.React`, what do you think?

MangelMaxime:
`Fable.Elmish.React` is supporting both at the same time:

Capture d’écran 2019-07-03 à 13 41 23

MangelMaxime:
Just to be more precise, `Fable.Elmish.React` use:

- `react-dom` API exposed from `Fable.React`: [source](https://github.com/elmish/react/blob/master/src/react.fs#L30)
- `react-native` API by importing it locally via interop: [source](https://github.com/elmish/react/blob/master/src/react-native.fs#L33)

Zaid-Ajaj:
Yeah I just read the source code as well, it is a real mess 😓 there should be a separate Elmish package for react native applications, the hierarchies become the following:

### Web Applications
```
Fable.Elmish.React
  |
  | -- Fable.ReactDom (react-dom >= 16.8.0)
         |
         | -- Fable.React (react >= 16.8.0)
```
### React Native Applications
```
Fable.Elmish.ReactNative
  |
  | -- Fable.ReactNative (react-native >= 0.57)
         |
         | -- Fable.React  (react >= 16.8.0)
```

alfonsogarciacaro:
I think this is a good idea but it's a breaking change for people depending on Fable.React without Elmish, we should release it as a new major version.

MangelMaxime:
> I think this is a good idea but it's a breaking change for people depending on Fable.React without Elmish, we should release it as a new major version.

Sure, I will prepare the packages and release them.

MangelMaxime:
I am closing in favour of #171

PR #171, opened 2019-07-24
WIP: Repo split

MangelMaxime:
Follow up on https://github.com/fable-compiler/fable-react/pull/169

The package split of `Fable.React` is ready, there is already a beta version of [Fable.React.Dom](https://github.com/fable-compiler/fable-react-dom) published.

I am waiting for a solution to [Fable.Elmish.React](https://github.com/elmish/react/issues/43) split before realising everything to stable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -258,3 +258,4 @@ packages/
 .idea/
 *.sln.iml
 *.orig
+.ionide/
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,9 @@
+### 6.0.0-beta-001
+
+- Add `forwardRef` to `IReactExports` (by @Luiz-Monad)
+- Remove `react-dom` API from this package
+- Add npm dependency metadata to work with Femto (by @Zaid-Ajaj)
+
 ### 5.2.6
 
 - Fix #167: Add `withKey` argument to `FunctionComponent.Of`
--- a/package-lock.json
+++ b/package-lock.json
@@ -298,9 +298,9 @@
       }
     },
     "lodash": {
-      "version": "4.17.11",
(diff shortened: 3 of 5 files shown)


Zaid-Ajaj:
This doesn't look ready!? because I see a lot of `react-dom` specific stuff still in here like Fable.React.Standard.fs, Props, Extensions etc... Now if Fable.ReactNative depends on Fable.React, they will have access to `div` and `h1` but they can't use them because they are `react-dom` specific..

MangelMaxime:
Ah yes, indeed.

I thought that react-native was accepting the same components as the one from the browser my bad.

I will update the PR.

Zaid-Ajaj:
I gave this split a try locally but it was soo messy because of the SSR stuff everywhere using compiler directives to generate html elements on the server, but now these stuff have to go to `fable-react-dom` or their own library?!

alfonsogarciacaro:
@MangelMaxime IIRC we never did the react/react-dom split at the end. Should we try to do it now that we are also splitting the types and the helpers?

MangelMaxime:
@alfonsogarciacaro Because we are already doing a breaking changes with Fable.React.Types, indeed I think it would be a good opportunity to split the packages.

alfonsogarciacaro:
What was the main motivation to remove react-dom btw? For cleanness or to avoid Femto using warnings in projects that don't need react-dom (like react native projects)?

MangelMaxime:
I think it was to have a clean binding in term of dependencies.

For example, native project doesn't need react-dom and it would probably also make it cleaner which package is responsible for what. As before everything was mixed together between F# specific code, bindings for the React.Dom or for React itself

alfonsogarciacaro:
Superseded by #234

Issue #150, opened 2019-04-01
JustifyContent should be made type safe just like AlignItems


MangelMaxime:

jannesiera:
It seems like a lot of `CSSProp`s are typed by an `obj` or a `string` where they could benefit from more type safety. Is there a reason why they're not typed by a DU, for example?

MangelMaxime:
It's because of `Fable.React` is one of the older libraries and at the time, we wanted to have a thin layer on top of React. So using `obj` make it easy to do that because the user can write `center`, `2px`, `calc(100vh - 3em)`, etc. 

We are also a small team of core maintainer and can't do all the evolution ourself :)

Because we accepted making some of the properties type-safe, we would accept a PR converting more properties to be type-safe :)

Luiz-Monad:
Is someone working on this?

Luiz-Monad:
Thinking better, that would be breaking change. https://github.com/fable-compiler/fable-react/issues/180

Issue #174, opened 2019-08-21
Cannot pattern match Style


halcwb:
Why can this code not be compiled to js?

```
let combineStyles st1 st2 =
    match st1, st2 with
    | Style p1, Style p2 -> p2 |> List.append p1 |> Style 
    | _ -> [] |> Style

```

I get:

> Module Error (from ./node_modules/fable-loader/index.js):
> ... error FSHARP: The pattern discriminator 'Style' is not defined. (code 39)
> 

Is this because of the ICSSProp vs CSSProp issue?

MangelMaxime:
Because when compiling against Fable `Style` isn't a DU case.

[Source](https://github.com/fable-compiler/fable-react/blob/master/src/Fable.React.Props.fs#L876-L882)

We need to do that, in order to convert the `Style` arguments into a JavaScript object as expected by React.

halcwb:
Thanks, so there isn't a way for what I want to achieve?

MangelMaxime:
Well, there are ways to achieve what you want but it relies on dynamic typing.

```fs
open Fable.Core
open Fable.Core.JsInterop
open Fable.React
open Fable.React.Props

[<Emit("Object.assign({}, $0, $1)")>]
let combineObjects (obj1 : obj) (obj2 : obj) = jsNative

let (|IsStyle|_|) (value : HTMLAttr) =
    if JS.Array.isArray(box value) then
        let _, style = unbox value
        Some (unbox<CSSProp list> style)
    else
        None

let combineStyles st1 st2 =
    match st1, st2 with
    | IsStyle p1, IsStyle p2 ->
        // p1 and p2 have already been converted to an JObject by Style helpers
        // some we need to combine both objects
        !! ("style", unbox combineObjects p1 p2)
    | _ -> 
        Style [ ]

let shouldReturnEmptyStyle = 
    combineStyles 
        (Disabled true) 
        (Style [ Height 20 ]) 

let shouldAppendBothStyles = 
    combineStyles 
        (Style [ Width 20 ]) 
        (Style [ Height 20 ]) 

JS.console.log(shouldReturnEmptyStyle)
JS.console.log(shouldAppendBothStyles)
```

Note that this code will only work against Fable runtime. If you are running it against SSR (.Net runtime) `Style` is indeed, in this case, a DUs Case. If you need to support both runtimes you can use compiler directives.

I am closing this issue, as the current behaviour is "normal" because of what React is expecting us to feed him with.

[...]

PR #175, opened 2019-08-24
Add type bindings for overflow, overflow-x and overflow-y options

jannesiera:
To get my feet wet regarding the strengthening of the[ type-safety of the CSS dsl](https://github.com/fable-compiler/fable-react/issues/150#issuecomment-524199843).

I've added options for overflow and its cousins overflow-x and overflow-y, accoding to the [spec](https://drafts.csswg.org/css-overflow-3/#propdef-overflow). All comments on the options are directly copied from their description in the spec.

### DIFF
--- a/src/Fable.React.Props.fs
+++ b/src/Fable.React.Props.fs
@@ -679,6 +679,164 @@ type BackfaceVisibilityOptions =
     | Inherit
     /// If the cascaded value of a property is the unset keyword, then if it is an inherited property, this is treated as inherit, and if it is not, this is treated as initial. This keyword effectively erases all declared values occurring earlier in the cascade, correctly inheriting or not as appropriate for the property (or all longhands of a shorthand).
     | Unset
+    
+[<StringEnum; RequireQualifiedAccess>]
+/// https://developer.mozilla.org/en-US/docs/Web/CSS/box-align
+type BoxAlignOptions =
+    | Start
+    | Center
+    | End
+    | Baseline
+    | Stretch
+    
+    /// Sets this property to its default value. Read more about [CSS Initial value](https://www.w3schools.com/CSSref/css_initial.asp)
+    | Initial
+    /// Inherits this property from its parent element. Read about [CSS Inherit](https://www.w3schools.com/CSSref/css_inherit.asp)
+    | Inherit
+    /// If the cascaded value of a property is the unset keyword, then if it is an inherited property, this is treated as inherit, and if it is not, this is treated as initial. This keyword effectively erases all declared values occurring earlier in the cascade, correctly inheriting or not as appropriate for the property (or all longhands of a shorthand).
+    | Unset
+    
+[<StringEnum; RequireQualifiedAccess>]
+type BoxDecorationBreakOptions =
+    /// Each box fragment is independently wrapped with the border, padding, and margin. The border-radius and border-image and box-shadow, if any, are applied to each fragment independently. The background is drawn independently in each fragment of the element. A no-repeat background image will thus be rendered once in each fragment of the element.
+    | Clone
+    /// The effect is as though the element were rendered with no breaks present, and then sliced by the breaks afterward: no border and no padding are inserted at a break; no box-shadow is drawn at a broken edge; and backgrounds, border-radius, and the border-image are applied to the geometry of the whole box as if it were unbroken.
+    | Slice
+    
+    /// Sets this property to its default value. Read more about [CSS Initial value](https://www.w3schools.com/CSSref/css_initial.asp)
+    | Initial
+    /// Inherits this property from its parent element. Read about [CSS Inherit](https://www.w3schools.com/CSSref/css_inherit.asp)
+    | Inherit
+    /// If the cascaded value of a property is the unset keyword, then if it is an inherited property, this is treated as inherit, and if it is not, this is treated as initial. This keyword effectively erases all declared values occurring earlier in the cascade, correctly inheriting or not as appropriate for the property (or all longhands of a shorthand).
+    | Unset
+
+[<StringEnum; RequireQualifiedAccess>]
+/// https://developer.mozilla.org/en-US/docs/Web/CSS/box-direction
+type BoxDirectionOptions =
+    /// The box lays out its contents from the start (the left or top edge).
+    | Normal
+    /// The box lays out its contents from the end (the right or bottom edge).
+    | Reverse
+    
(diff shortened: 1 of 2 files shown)


MangelMaxime:
Hum, we are facing a problem here.

`overflow: hidden visible` is a valid CSS value.

> The overflow property is specified as one or two keywords chosen from the list of values below. If two keywords are specified, the first applies to overflow-x and the second to overflow-y. Otherwise, both overflow-x and overflow-y are set to the same value.

jannesiera:
Seems like I was being a bit too eager here :).

Would something like this work?

```
    static member Overflow (overflow: OverflowOptions, ?overflowY: OverflowOptions) =
        let values =
            match overflowY with
            | Some value -> [ overflow; value ]
            | None -> [ overflow ]
        unbox ("overflow", keyValueList CaseRules.LowerFirst values)
```

MangelMaxime:
No, because it needs to return a `string` not an `Object`

*This need to be tested*

```fs
    static member Overflow (overflow: OverflowOptions, ?overflowY: OverflowOptions) =
        unbox ("overflow",  unbox overflow + " " + unbox overflowY)
```

We can `unbox` `OverflowOptions` values to `string` type because `OverflowOptions` is marked with `StringEnum` and so they indeed are `string` at runtime.

jannesiera:
Right! I'll go ahead and test this one then.

jannesiera:
I've tested these changes in a consumer project and they work.

Returning the unboxed value directly probably wasn't a good idea, and wouldn't work for the SSR code. So I ended up returning a `CSSProp.Custom` with the value.

jannesiera:
I was too quick with my test and noticed that it actually didn't work when only passing a single argument (leaving out the optional value). Not unboxing the optional directly fixed that issue.

jannesiera:
I believe these changes could be merged.

I'm planning to do more of these, and hopefully cover all the properties. It might be good to go step by step.

MangelMaxime:
I am waiting @alfonsogarciacaro feedbacks.

Personally, I think if we accept changes like these we need to do all of them in one major release (can be stored in a separate branch and contribute into several PRs). Otherwise, each time someone upgrade `Fable.React` he will have breaking changes.

1. It's annoying to have breaking on each minor updates
2. Make it feels like unstable

jannesiera:
Added DUs for various other CSSProps.

src/Fable.React.Props.fs
```diff
@@ -872,6 +1087,11 @@ type CSSProp =
     | Zoom of obj
     /// If you are searching for a way to provide a value not supported by this DSL then use something like: CSSProp.Custom ("align-content", "center")
     | [<Erase>] Custom of string * obj
+    static member Overflow (overflow: OverflowOptions, ?overflowY: OverflowOptions) =
+        match overflowY with
+        | Some value -> CSSProp.Custom ("overflow",  unbox overflow + " " + unbox value)
+        | None -> CSSProp.Custom ("overflow",  unbox overflow)
```
alfonsogarciacaro: Hmm, not sure if this will work on .NET side for SSR, but this makes me realize we may also have a problem with other StringEnum options. I need to look into that.

alfonsogarciacaro:
I will merge this shortly and fix the SSR issue myself because this has made me realize we have an issue with the other CSS props.

However, as it's a breaking change in principle we should release it as a new major version. @MangelMaxime What's the situation with the react/react-dom split? Has something be decided for elmish/react?

MangelMaxime:
> However, as it's a breaking change in principle we should release it as a new major version. 

I agree but only if all the options have been converted to the typed way otherwise it will feel hacky. If not, we can just release a fix for the SSR code.

> What's the situation with the react/react-dom split? Has something be decided for elmish/react?

I need to take the time to do the split because it was a bit harder than planned :) You can ignore this situation, I will adapt the repos when ready. I am focusing on Nacara right now to solve several issues before starting another task.

jannesiera:
> > However, as it's a breaking change in principle we should release it as a new major version.
> 
> I agree but only if all the options have been converted to the typed way otherwise it will feel hacky. If not, we can just release a fix for the SSR code.

I concur.

I'd really like to cover all the props. There's a lot of them so it's going to take a lot of time. I'm starting with the simpler ones that can be turned in a DU with a `StringEnum` annotation. Afterwards there will be a lot of them which will need a similar treatment as the `overflow` prop because they can receive optional parameters, etc.

I'll also do a proposal to treat pixels/percentages/length etc as a seperate type to have them type-safe and compliant with the [css spec](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Types). A proposal is not ready yet, and I'd be happy to hear your opinions.

MangelMaxime:
No worry @jannesiera my goal isn't to rush you or someone to do it for tomorrow :)

But, I wasn't comfortable with shipping current version of Fable.React with various style (some type safe and some obj) so I would like to no redo it.

Especially, when your PR and the code for fixing SSR are not linked together. :)

alfonsogarciacaro:
I'm thinking about the issues with SSR: most of them happen because the special/hacky features of Fable like `keyValueList` or `StringEnum` (causing also situations like #174). These were ones of the few hacks surviving from Fable 1 to allow the compiler optimization that converts the props list to a JS object directly at compile time (when possible)

However, as this is causing many little issues when sharing the code with .NET maybe it's better to discard the hacks and use more standard code. This may bring a little runtime overhead but it shouldn't be noticeable. What do you think?

alfonsogarciacaro:
About the units, it's a good idea. Though as we're using unit types, we would have to add extra cases (like HeightPx, HeightPercentage...) or mask static members as union cases (another hack).

Also, @Zaid-Ajaj Feliz project is already exploring this direction. Maybe we can wait for feedback from users of that library before making similar changes here.

MangelMaxime:
> This may bring a little runtime overhead but it shouldn't be noticeable.

Compare to today, this means that we will iterate over each list element one more time in order to check if we need to convert the `Style` element.

As a side note, if we change the `Style` representation, this has to be done as a major version because it can break existing library I think.

Is SSR really broken? Because I am surprised that no one complained yet about it.

alfonsogarciacaro:
TBH, I don't know. I don't use SSR myself and probably not many people use it yet. If there's a problem, it's probably a few CSS rules not working correctly and they may be overriden by React on the frontend. That's maybe the reason nobody has noticed it yet.

alfonsogarciacaro:
Ok, I'll merge this PR so I don't forget and I'll try to see what can I do about the SSR thing :)

Issue #140, opened 2019-01-31
API for React Hooks


alfonsogarciacaro:
Seems React Hooks will be shipping in stable form soon and the API doesn't seem to have changed much. We will have to add bindings for them soon and as usual we've two alternatives: a) match the native API as close as possible, or b) be a bit imaginative and try to make something more F# friendly. I've been playing with the two.

### Matching native API

```fsharp
type Hooks =
    [<Import("useState", from="react")>]
    static member useState<'T> (initialState: 'T): 'T * ('T->unit) = jsNative

    [<Import("useEffect", from="react")>]
    static member useEffect<'T> (effect: unit -> (unit->unit) option, ?checkValues: obj[]): unit = jsNative

// Usage
type Props =
    { MessageFormat: string }

let MyComponent (props: Props) =
    let state, setState = Hooks.useState("")

    Hooks.useEffect(fun () ->
        printfn "New render: %s" state
        None)

    div [ Class "main-container" ]
        [ input [ Class "input"
                  Value state
                  OnChange (fun ev -> ev.target?value |> setState) ]
          span [ ]
            [ str (String.Format(props.MessageFormat, state)) ] ]

ofFunction MyComponent { MessageFormat = "Hello, {0}!" } []
|> mountById "app"
```

[...]

PR #176, opened 2019-08-24
Use reducer

nojaf:
Adding `use-reducer` in the style of `use-state`.
https://reactjs.org/docs/hooks-reference.html#usereducer
Related to #140


alfonsogarciacaro:
Thanks!

PR #178, opened 2019-09-02
void tags should contain a space to be XHTML1.0 compliant

forki:


forki:
This should render `<hr />` instead of `<hr/>`

alfonsogarciacaro:
Thanks! Just in time for a new release!

PR #181, opened 2019-10-10
WIP: Missing hyphens for typed css props

SCullman:
I noticed that `WhiteSpace WhiteSpaceOptions.PreLine` was not working. 
If there is not an option missing, Compiled Name has to be set for a lot of props.

### DIFF
--- a/src/Fable.React.Props.fs
+++ b/src/Fable.React.Props.fs
@@ -523,11 +523,11 @@ type WhiteSpaceOptions =
     /// Like normal, this value collapses white space; but like pre, it does not allow wrapping.
     | Nowrap
     /// Like pre, this value preserves white space; but like normal, it allows wrapping.
-    | PreWrap
+    | [<CompiledName "pre-wrap">] PreWrap
     /// The behavior is identical to that of pre-wrap, except that: 1. Any sequence of preserved white space always takes up space, including at the end of the line. 2. A line breaking opportunity exists after every preserved white space character, including between white space characters.
     | BreakSpaces
     /// Like normal, this value collapses consecutive spaces and allows wrapping, but preserves segment breaks in the source as forced line breaks.
-    | PreLine
+    | [<CompiledName "pre-line">] PreLine
 
     /// Sets this property to its default value. Read more about [CSS Initial value](https://www.w3schools.com/CSSref/css_initial.asp)
     | Initial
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
Ah, good one! I'm surprised we didn't catch this earlier :confused: I remember now we were talking about adding `SnakeCase` to the [CaseRules options](https://github.com/fable-compiler/Fable/blob/master/src/Fable.Core/Fable.Core.Types.fs#L5-L7), but apparently I forgot to do it at the end 😅 

You mean this change is needed in other options too? That's why the PR is WIP?

SCullman:
> You mean this change is needed in other options too?

Looks like the tip of an iceberg

> That's why the PR is WIP?

I hoped for a kind of snake-case option,  because on first sight it seemed that nearly any PascalCase word needs an hyphen.

Thank you for answering immediately. Tomorrow I will have a longer ride on a train, good chance to run through all props...

alfonsogarciacaro:
Hmm, I think you're right, we should use this opportunity to implement SnakeCase in the CaseRules. This would mean users need to update the fable-compiler version which unfortunately there's no way to force at the moment from a library. But given the CSS rules are already broken, I guess the worst thing that can happen is users won't get the fix until they update fable-compiler.

SCullman:
That would be great. There are more than 300 CSSProps. In checked about 50%. Either SnakeCase fits, or the props are unknown to Chrome :-).
So I appreciate if you implement SnakeCase.
In the mean time, may this state be merged as minor release? It would help in my project.

alfonsogarciacaro:
Sure thing! Published as 5.3.3

SCullman:
Whow, thanks!

src/Fable.React.Props.fs
```diff
@@ -893,9 +893,9 @@ type UserSelectOptions =
     | Unset
 
 type CSSProp =
-    | AlignContent of AlignContentOptions
-    | AlignItems of AlignItemsOptions
-    | AlignSelf of AlignSelfOptions
+    | [<CompiledName "align-content">] AlignContent of AlignContentOptions
+    | [<CompiledName "align-items">] AlignItems of AlignItemsOptions
```
forki: React seems to be disagreeing with this one. https://github.com/fable-compiler/fable-react/issues/182

SCullman: Yes, I should not have touched the CssProps. The CompiledName is only required for Options.

PR #183, opened 2019-10-14
Removed Hyphen from CssProp

### LINKED ISSUE #182: "AlignItems AlignItemsOptions.FlexStart" broken in latest version
The follwoing code used to work:

image

Unfortunately lastet fable-react broke it:

image

SCullman:
CssProps are, unlike their options, no string enums. They MUST not have any hyphen.
fixes #182

### DIFF
--- a/src/Fable.React.Props.fs
+++ b/src/Fable.React.Props.fs
@@ -893,9 +893,9 @@ type UserSelectOptions =
     | Unset
 
 type CSSProp =
-    | [<CompiledName "align-content">] AlignContent of AlignContentOptions
-    | [<CompiledName "align-items">] AlignItems of AlignItemsOptions
-    | [<CompiledName "align-self">] AlignSelf of AlignSelfOptions
+    | AlignContent of AlignContentOptions
+    | AlignItems of AlignItemsOptions
+    | AlignSelf of AlignSelfOptions
     | AlignmentAdjust of obj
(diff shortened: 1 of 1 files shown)


SCullman:
@alfonsogarciacaro , this means also there is actually no need for a SnakeCase option.

forki:
Thanks

Stefan Cullmann <notifications@github.com> schrieb am Mo., 14. Okt. 2019,
13:20:

MangelMaxime:
Released in 5.3.4.

I didn't test the code, so if there is still a bug please open an issue or send a PR ;)

alfonsogarciacaro:
I was about to include the remaining the [remaining CSS Prop options](https://github.com/fable-compiler/fable-react/commit/abbd431c471b89f486db613860f346a2cfe80391) in the new release, but you were faster than me @MangelMaxime ;)

Thanks for reporting @forki, this escaped to me in the previous PR.

@SCullman With the SnakeCase rule, we would be able to just write `StringEnum(CaseRules.SnakeCase)` for the CSS Prop options instead of having to manually add `CompiledName` in all multi-wod cases. It will be fixed anyways in the next version, but `SnakeCase` should be still be a valuable addition for the future.

forki:
thanks for quick fix and release. Works for me!

Am Mo., 14. Okt. 2019 um 13:32 Uhr schrieb Maxime Mangel <
notifications@github.com>:

PR #184, opened 2019-10-24
Add aria-XXX attributes

Krzysztof-Cieslak:
Source: w3 doc in https://www.w3.org/TR/wai-aria/#aria-atomic


alfonsogarciacaro:
Thanks! I'll push a new release :+1:

PR #185, opened 2020-01-16
Added JustifySelf used in Grid containers

Luiz-Monad:
Support for CSS JustifySelf  (https://developer.mozilla.org/en-US/docs/Web/CSS/justify-self)

It's still working draft, but we have JustifyContent anyway.
They'll probably move the grid spec to recommendation soon.


MangelMaxime:
Thank you it has been released in version 5.3.6

PR #187, opened 2020-03-24
Fixes webpack ReferenceError for SSRSample

Kethet:
Attempting to launch the SSRSample gives `ReferenceError: webpack is not defined`.


MangelMaxime:
Good catch :) Thank you

PR #193, opened 2020-04-24
Fix notation for SVG attributes SSR

mvsmal:
This PR fixes notation in SSR for some of the predefined SVG attributes. The changes are made according to this reference: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute


MangelMaxime:
Thank you it has been released in version 6.2.0

mvsmal:
hi @MangelMaxime ! Thanks a lot for the quick merge an release! Would it be possible to release this fix also in version 5.x? The problem is that we cannot even work it around since `SVGAttr.Custom` is rendered with `slugKey` and we are not ready to update to v.6 yet.

MangelMaxime:
Version 5.4.0 has been released with the fix in it

PR #195, opened 2020-05-08
Fix #194

### LINKED ISSUE #194: Update from v5 to v6 breaks application
I've updated my application to v6 of Fable.React and while everything compiles fine, I am now getting a blank page with this error:

```
Uncaught TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
    at eval (webpack-internal:///./.fable/fable-library.2.8.4/Seq.js:692)
    at Object.eval (webpack-internal:///./.fable/fable-library.2.8.4/Seq.js:385)
    at Object.eval (webpack-internal:///./.fable/fable-library.2.8.4/Seq.js:385)
    at fold (webpack-internal:///./.fable/fable-library.2.8.4/Seq.js:551)
    at ofSeq (webpack-internal:///./.fable/fable-library.2.8.4/List.js:277)
    at attributeTags (webpack-internal:///./src/Client/Common.fs:530)
    at prompt (webpack-internal:///./src/Client/Common.fs:538)
    at renderWithHooks (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:14803)
    at mountIndeterminateComponent (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:17482)
    at beginWork (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:18596)
```

`prompt` is a function component defined in a shared module like this:

```fsharp
let memo name r = FunctionComponent.Of (memoizeWith = equalsButFunctions, displayName = name, render = r)

let prompt = memo "prompt" (fun (x: {| some props |}) ->
    Columns.columns [] [
        attributeTags blah
    ])
```

and used like `div [] [ prompt {| prop = "" |} ]`. It's worth mentioning though that it is not rendered anywhere near the entry point of the application, so I don't know why the inside of its render function should be called and throw errors on page load.

Reverting back to v5 (and an earlier version of Feliz, the last one that depends on v5, and which I don't use, but is a transitive dependency of Feliz.Router) causes everything to go back to normal.

alfonsogarciacaro:
Ok, this is my new attempt to fix the `FunctionComponent.Of` API #188 #194. To make sure users are always passing the same render function to each `FunctionComponent.Of` I'm storing a stringified version of the function in the cache and compare it with newly passed render arguments. Seems to work but please someone give it a try to double check. I'm doing this only in Debug mode to avoid the overhead in production. I hope it's enough but let me know what you think.

This PR also updates npm dependencies to remove the security warnings (there's still one about yargs-parser when running `npm audit` that I couldn't solve).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3,28 +3,40 @@
   "lockfileVersion": 1,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.5.5",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
-      "integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
+      "version": "7.8.3",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
+      "integrity": "sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==",
       "requires": {
-        "@babel/highlight": "^7.0.0"
+        "@babel/highlight": "^7.8.3"
+      }
+    },
+    "@babel/compat-data": {
+      "version": "7.9.6",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.9.6.tgz",
+      "integrity": "sha512-5QPTrNen2bm7RBc7dsOmcA5hbrS4O2Vhmk5XOL4zWW/zD/hV0iinpefDlkm+tBBy8kDtFaaeEvmAqt+nURAV2g==",
+      "requires": {
+        "browserslist": "^4.11.1",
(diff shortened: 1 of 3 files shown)


MangelMaxime:
Looking at the code, I think it should work.

I guess we should try to release a beta version this time and ask people having the issue with the upgrade to comment if that helps them or not. Either by fixing the problem or helping troubleshoot the problem :)

GBirkel:
I think this fixes a rendering issue we’re having with our recent merge at Demetrix...

alfonsogarciacaro:
Hi @GBirkel! Please note that unfortunately this is not a "fix", just a way to fail fast and warn the user. If you're having problems maybe it's safer to go back to Fable.React 5.x.

@MangelMaxime Should we unlist the 6.x package if users are having issues?

MangelMaxime:
@alfonsogarciacaro I don't know in general I don't like the unlisting package but if that avoid problem perhaps we should.

Also, now it has been out for several days/weeks so perhaps it is too late?

Release this new fast failing would help people make the transition should be enough in theory no?

forki:
With standard nuget unlisted packages are still installed if they are the
latest. It's a bug that they won't fix. So you always need to Unlist + push
a new one

Maxime Mangel <notifications@github.com> schrieb am Sa., 16. Mai 2020,
18:47:

MangelMaxime:
@alfonsogarciacaro Should we release this new version?

PR #207, opened 2020-10-15
ReactComponent Attribute

alfonsogarciacaro:
This PR supersedes #196 and uses the plugin system from Nagareyama instead (see also https://github.com/fable-compiler/Fable/pull/2208). The code for the plugin is a bit more convoluted because the whole Fable.AST is now exposed. I haven't published the internal helpers for AST manipulation because the API wasn't designed for public consumption and it's a bit messy. For now I just added some simple helpers in the `AstUtil`module.

The workflow is as follows. I hope it's good enough but we can try to improve it if needed:

- Plugins must go in dll assemblies that are not compiled by Fable. Basically this means they go in a separate package (you can still reference it from your Fable library) and **not** include the sources (so a tag [like this](https://github.com/fable-compiler/fable-react/blob/106c7f6eaedef527dd5536b96f4bd22e2a5014db/src/Fable.React.fsproj#L23-L25) is not necessary).
- This assembly must reference the Fable.AST package.
- The plugin is just an attribute that inherits one of the `Fable.PluginAttribute` children (right now, `MemberDeclarationPlugin` and `CallPlugin`).
- That's basically it, Fable will automatically detect the plugin when loading the .dll assembly.

### Development

During development you don't want to republish the package to test every change. In that case, use a project reference (see TodoMVC.fsproj in this pr) together with the `--exclude` option, which instructs F# to ignore the sources from referenced projects and use the compiled .dll assemblies instead. For this, make sure you build the project containing the plugin before starting Fable. The steps would be:

```
dotnet build my/plugin.project
dotnet fable my/test.sample --exclude plugin.project
```

You can see a more complex example of this in the new npm `todomvc-start` script: https://github.com/fable-compiler/fable-react/blob/9bc42fb37aa77695b0d2cca3455b1d41ab633d56/package.json#L5

@Zaid-Ajaj I hope this is a good guide to start hacking with the AST. Please give it a try when you have a moment and let me know how it goes.

### DIFF
--- a/plugins/Fable.React.Plugins.fs
+++ b/plugins/Fable.React.Plugins.fs
@@ -24,8 +24,8 @@ module internal AstUtil =
         Fable.StringConstant x |> makeValue None
 
     let makeImport selector path =
-        Fable.Import({ Selector = makeStrConst selector
-                       Path = makeStrConst path
+        Fable.Import({ Selector = selector
+                       Path = path
                        IsCompilerGenerated = true }, Fable.Any, None)        
 
     let makeCall callee args =
@@ -33,6 +33,7 @@ module internal AstUtil =
             { ThisArg = None
               Args = args
               SignatureArgTypes = []
+              CallMemberInfo = None
               HasSpread = false
               IsJsConstructor = false }
         Fable.Call(callee, callInfo, Fable.Any, None)
@@ -71,14 +72,18 @@ module internal AstUtil =
             Body = v
             UsedNames = Set.empty
             Info = MemberInfo(isValue=true)
+            ExportDefault = false
         }
 
     let objExpr kvs =
         Fable.ObjectExpr(List.map objValue kvs, Fable.Any, None)
 
 
-type ReactComponentAttribute() =
+type ReactComponentAttribute(exportDefault: bool) =
     inherit MemberDeclarationPluginAttribute()
+
+    new() = ReactComponentAttribute(exportDefault=false)
+
     override _.FableMinimumVersion = "3.0"
 
     override _.TransformCall(helper, memb, expr) =
@@ -109,4 +114,4 @@ type ReactComponentAttribute() =
                 |> fun bindings -> Fable.Let(bindings, decl.Body)
 
             // TODO: check if decl.Name is uppercase
-            { decl with Args = [propsArg]; Body = body }
+            { decl with Args = [propsArg]; Body = body; ExportDefault = exportDefault }
--- a/plugins/Fable.React.Plugins.fs
+++ b/plugins/Fable.React.Plugins.fs
@@ -12,9 +12,7 @@ module internal AstUtil =
     let makeIdent name: Fable.Ident =
         { Name = name
           Type = Fable.Any
-          // It's important to set this to false
-          // so Fable doesn't optimize the binding
-          IsCompilerGenerated = false
+          IsCompilerGenerated = true
           IsThisArgument = false
           IsMutable = false
           Range = None }
@@ -82,27 +80,33 @@ module internal AstUtil =
 type ReactComponentAttribute() =
     inherit MemberDeclarationPluginAttribute()
     override _.FableMinimumVersion = "3.0"
+
+    override _.TransformCall(helper, memb, expr) =
+        let membArgs = memb.CurriedParameterGroups |> List.concat
+        match expr with
+        | Fable.Call(callee, info, t, r) when List.length membArgs = List.length info.Args ->
+            let propsObj =
+                List.zip membArgs info.Args
+                |> List.choose (fun (arg, expr) -> arg.Name |> Option.map (fun k -> k, expr))
+                |> AstUtil.objExpr
+            // TODO: Detect `children` arg
+            AstUtil.makeCall (AstUtil.makeImport "createElement" "react") [callee; propsObj]
+        | _ -> expr
+
     override _.Transform(logger, decl) =
         if decl.Info.IsValue || decl.Info.IsGetter || decl.Info.IsSetter then
             logger.LogWarning("Expecting a function for ReactComponent")
             decl
         else
             let propsArg = AstUtil.makeIdent "$props"
-            let renderIdent = AstUtil.cleanFullDisplayName decl.FullDisplayName |> AstUtil.makeIdent
-            let renderBody =
+            let body =
                 ([], decl.Args) ||> List.fold (fun bindings arg ->
+                    // TODO: Skip getter if arg is `key`
                     let getterKind = Fable.ByKey(Fable.ExprKey(AstUtil.makeStrConst arg.DisplayName))
                     let getter = Fable.Get(Fable.IdentExpr propsArg, getterKind, Fable.Any, None)
                     (arg, getter)::bindings)
                 |> List.rev
                 |> fun bindings -> Fable.Let(bindings, decl.Body)
 
-            let propsObj =
-                decl.Args
-                |> List.map (fun arg -> arg.DisplayName, Fable.IdentExpr arg)
-                |> AstUtil.objExpr
-            let body =
-                let createEl = AstUtil.makeCall (AstUtil.makeImport "createElement" "react") [Fable.IdentExpr renderIdent; propsObj]
-                Fable.Let([renderIdent, Fable.Delegate([propsArg], renderBody, None)],
-                    Fable.Delegate(decl.Args, createEl, None))
-            { decl with Args = []; Body = body; Info = AstUtil.MemberInfo(decl.Info, isValue=true) }
+            // TODO: check if decl.Name is uppercase
+            { decl with Args = [propsArg]; Body = body }
--- a/plugins/Fable.React.Plugins.fs
+++ b/plugins/Fable.React.Plugins.fs
@@ -1,0 +1,108 @@
+namespace Fable.React
+
+open Fable
+open Fable.AST
+
+module internal AstUtil =
+    open System.Text.RegularExpressions
+
+    let cleanFullDisplayName str =
+        Regex.Replace(str, @"`\d+", "").Replace(".", "_")
+
+    let makeIdent name: Fable.Ident =
+        { Name = name
+          Type = Fable.Any
+          // It's important to set this to false
+          // so Fable doesn't optimize the binding
+          IsCompilerGenerated = false
+          IsThisArgument = false
+          IsMutable = false
+          Range = None }
+
+    let makeValue r value =
+        Fable.Value(value, r)
+
+    let makeStrConst (x: string) =
+        Fable.StringConstant x |> makeValue None
+
+    let makeImport selector path =
+        Fable.Import({ Selector = makeStrConst selector
+                       Path = makeStrConst path
+                       IsCompilerGenerated = true }, Fable.Any, None)        
+
+    let makeCall callee args =
+        let callInfo: Fable.CallInfo =
+            { ThisArg = None
+              Args = args
+              SignatureArgTypes = []
+              HasSpread = false
+              IsJsConstructor = false }
+        Fable.Call(callee, callInfo, Fable.Any, None)
+
+    type MemberInfo(?info: Fable.MemberInfo,
+                    ?isValue: bool) =
+
+        let infoOr f v =
+            match info with
+            | Some i -> f i
+            | None -> v
+
+        let argOrInfoOr arg f v =
+            match arg, info with
+            | Some arg, _ -> arg
+            | None, Some i -> f i
+            | None, None -> v
+
+        interface Fable.MemberInfo with
+            member _.IsValue = argOrInfoOr isValue (fun i -> i.IsValue) false
+            member _.Attributes = infoOr (fun i -> i.Attributes) Seq.empty
+            member _.HasSpread = infoOr (fun i -> i.HasSpread) false
+            member _.IsPublic = infoOr (fun i -> i.IsPublic) true
+            member _.IsInstance = infoOr (fun i -> i.IsInstance) true
+            member _.IsMutable = infoOr (fun i -> i.IsMutable) false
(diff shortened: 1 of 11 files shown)


Zaid-Ajaj:
Awesome stuff @alfonsogarciacaro, I will be diving into it very soon, probably after this weekend. I have one question about 

> Plugins must go in dll assemblies that are not compiled by Fable. Basically this means they go in a separate package (you can still reference it from your Fable library) and not include the sources (so a tag like this is not necessary).

Does that mean I cannot implement a plugin in the library itself (say Feliz) but instead would _require_ to implement a separate package (like Feliz.Plugin) where the plugin is implemented?

alfonsogarciacaro:
Yes, sorry. I did try putting the plugin in the library itself at first but had issues when Fable tried to compile the plugin code. Putting the plugin into a separate assembly/package has some advantages (clear distinction between plugin and to-be-js-compiled code, Fable compilation has less work to do, you can use .NET APIs), but if it becomes too cumbersome we can try some work around, like including the Fable.AST sources to make it Fable-compatible or erase the plugin code somehow. Though in that case we may have some challenges to locate the .dll containing the plugin, particularly during development.

Zaid-Ajaj:
> Yes, sorry. I did try putting the plugin in the library itself at first but had issues when Fable tried to compile the plugin code. Putting the plugin into a separate assembly/package has some advantages (clear distinction between plugin and to-be-js-compiled code, Fable compilation has less work to do, you can use .NET APIs)

Doesn't adding `!FABLE_COMPILER` is enough to tell Fable not to compile the plugin code? At the end of the day, when we are publishing the nuget packages, we are not defining the `FABLE_COMPILER` constant which means:
 - Dlls will contain the IL code of the plugin
 - The sources will contain the plugin code, except that Fable won't compile it

Another thing just to double check: consumers of libraries with plugins don't need to exclude the plugin from the compilation, correct? Only developers of the library need to use the `--exclude` option if I understand correctly

alfonsogarciacaro:
Forgot to comment about conditional compilation. Fable compilation has two phases: FCS compilation to generate the F# AST, and Fable compilation to translate the AST to Fable -> Babel (and now also to print the JS code). If you hide the attribute definition behind `!FABLE_COMPILER` then FCS will fail when using the attribute because the compiler doesn't know anything about it. You would have to hide only the transform code like:

```fsharp
type ReactComponentAttribute() =
    inherit MemberDeclarationPluginAttribute()
    override _.FableMinimumVersion = "3.0"
    override _.Transform(logger, decl) =
#if FABLE_COMPILER
       failwith "Not supported"
#else
       ...
#endif
```

And you also need to hide code used by the plugin like the `AstUtil` module so it can become complicated quite quickly.

> consumers of libraries with plugins don't need to exclude the plugin from the compilation

That's right. For consumers everything will be transparent. They don't even need to install `Feliz.Plugins` because it will be a transitive dependency of `Feliz`. Fable will automatically skip merging the sources of packages if it doesn't find a `fable` folder in the Nuget package. `--exclude` will only be need during development if you use project references (I'm thinking now of a way to automatically rebuild the dll in that case). We do the same when compiling Fable.Library or the Fable tests when referencing the Fable.Core project directly: https://github.com/fable-compiler/Fable/blob/3c51973fb030aa1f23c1a008d470c5d7cbcc611d/build.fsx#L81

If you use local Nuget packages when developing you shouldn't even need to use the `--exclude` option (only `--forcePkgs` to clear the cached sources every time).

Zaid-Ajaj:
Thanks a lot @alfonsogarciacaro for the thorough explanation 🙏 I will start working on the plugin as soon as time permits

alfonsogarciacaro:
`3.0.0-nagareyama-alpha-014` will automatically build the excluded project references, so the command has been simplified to:

```
cd Samples/TodoMVC
dotnet fable watch src --noCache --exclude Fable.React.Plugins --run webpack-dev-server
```

plugins/Fable.React.Plugins.fs
```diff
@@ -0,0 +1,108 @@
...
+                    (arg, getter)::bindings)
+                |> List.rev
+                |> fun bindings -> Fable.Let(bindings, decl.Body)
+
+            let propsObj =
+                decl.Args
+                |> List.map (fun arg -> arg.DisplayName, Fable.IdentExpr arg)
+                |> AstUtil.objExpr
+            let body =
+                let createEl = AstUtil.makeCall (AstUtil.makeImport "createElement" "react") [Fable.IdentExpr renderIdent; propsObj]
```
Zaid-Ajaj, line 105: @alfonsogarciacaro Although this works, combining the component definition with component creation will _most likely_ cause problems for fast-refresh plugin that provides live-reload capabilities for pure React apps. I need to test it first to make sure but I according the discussions I had with the people working on fast-refresh, I believe we will need to have two Fable plugins: one that rewrites the input arguments into `$prop` declaration _and_ another Fable plugin that rewrites call-site into a `createElement` call:
```fs
[<ReactComponent>]
let hello (name: string) = Html.div ("Hello, " + name)

// somewhere down the line
hello "world"
```
should compile to 
```js
export const Hello = ($props) => {
  const { name } = $props
  return Feliz_Html.div("Hello, " + name)
}

createElement(Hello, { name: "world" })
```
Do you think something like that is possible? 

> Notice the upper case transform, this is also something required by the fast-refresh plugin

alfonsogarciacaro: I see, thanks for letting me know. It should be possible with the CallPlugin. Let me give it a try.

Zaid-Ajaj: Bonus points if we could elminite `$props` altogether and instead spread the arguments when defining the component:
```js
export const Hello = ({ name }) => {
  return Feliz_Html.div("Hello, " + name)
}
```
I think there will also be some edge cases when someone uses `key` and `children` as input arguments where:
 - `key` can be provided on call-site but is not available from inside the component (show warnings when `key` is used?)
 - `children` provided on call-site should be _spread_ (like what `[<ParamList>]` does) as the third argument of `createElement`

But you can leave those for me to work on :wink: maybe I won't have to implement a Feliz specific plugin if this does the correct transforms and I would only implement Feliz specific transforms to simplify the generated javascript

alfonsogarciacaro: Hmm, that's gonna be tricky because we have nothing in Fable AST for object spread, only for spreading the last argument in call (we haven't added it to our Babel AST implementation yet either) and we cannot use `Emit` with the arguments. We would have to detect the pattern either in the Fable > Babel or printing phases but it may not be easy.

About `key` and `children` we should be able to deal with these differently to other arguments, I guess.

Zaid-Ajaj: > Hmm, that's gonna be tricky because we have nothing in Fable AST for object spread, only for spreading the last argument in call (we haven't added it to our Babel AST implementation yet either) and we cannot use Emit with the arguments. We would have to detect the pattern either in the Fable > Babel or printing phases but it may not be easy.

It's alright, it is just cosmetics but definitely a nice to have

alfonsogarciacaro: @Zaid-Ajaj Please check the [latest commit](https://github.com/fable-compiler/fable-react/pull/207/commits/6b81f1e8656fe222e5ff00e43ea65c3f1cc22e82). You need to update both fable tool and Fable.AST but now you can transform the call site too 👍

Zaid-Ajaj: @alfonsogarciacaro one question, how to determine that a `Fable.Type` is a record instance? Right now I have this:
```fs
let isRecord (fableType: Fable.Type) =
    match fableType with
    | Fable.Type.AnonymousRecordType _ -> true
    | _ -> false
```
I want to _not_ rewrite the props when the only input used is a single record instance or anonymous record but I am not sure how to check for records :/

I know it must be must something that has to do with `EntityRef` but I only get a `name` out of it and I cannot check whether that name is actually referring to a record type

alfonsogarciacaro: Ah, yes, sorry. I removed the direct references for entities in the AST (this is done to make serialization easier later). I was thinking we should pass the [Compiler object](https://github.com/fable-compiler/Fable/blob/f37a97e61bb67d0c921a1d3780cb361ff975d9fb/src/Fable.Transforms/Global/Compiler.fs#L54-L66) (or parts of it, as it currently has a dependency to FCS) through the plugin helper, and this is probably a good reason for that [as you can use the compiler to get the entity from the reference](https://github.com/fable-compiler/Fable/blob/f37a97e61bb67d0c921a1d3780cb361ff975d9fb/src/Fable.Transforms/Replacements.fs#L739-L744). I'll do it for the next Fable.AST/fable update 👍

Zaid-Ajaj: That would be great! thanks a lot :pray: 🙇 

I am making quite the progress with these plugins, very cool stuff :heart:
```

**Comment by nojaf** [comment]
Hey @alfonsogarciacaro, any news on this PR?
As I'm using the original Elmish DSL in all my projects and I'm interested in `react-refresh` I'm really looking forward to this plugin.
Any chances we can get an alpha/beta of this?

**Comment by alfonsogarciacaro** [comment]
Hi @nojaf! I was a bit concerned that if we publish two libraries with a similar `ReactComponent` attribute it could be a bit confusing for users, especially those mixing Fable.React and Feliz DSLs in their apps. So for now I was thinking to let users try @Zaid-Ajaj version and see how it works. In fact, you should be able to use Feliz.CompilerPlugins (which is already published) in your Fable.React apps without the Feliz dependency. Can you please give it a try and let us know if you have any problem?

**Comment by nojaf** [comment]
Hey @alfonsogarciacaro, that totally worked! Many thanks for pointing this out.

File: plugins/Fable.React.Plugins.fs [hunk 2]
```diff
@@ -79,6 +79,9 @@ module internal AstUtil =
         Fable.ObjectExpr(List.map objValue kvs, Fable.Any, None)
 
 
+[<assembly:ScanForPlugins>]
+do()
```
alfonsogarciacaro, line 83: I found a way to avoid having to scan all assemblies, so this is necessary now to tell Fable that this assembly in particular does contain plugins and should be scanned. Also, please put the plugins directly in a namespace and not inside a module.

plugins/Fable.React.Plugins.fs
```diff
@@ -98,7 +101,7 @@ type ReactComponentAttribute(exportDefault: bool) =
             AstUtil.makeCall (AstUtil.makeImport "createElement" "react") [callee; propsObj]
         | _ -> expr
 
-    override _.Transform(logger, decl) =
+    override _.Transform(logger, _file, decl) =
```
alfonsogarciacaro, line 104: I added an extra argument with a handle to the whole file Fable AST in case it's needed later for https://github.com/fable-compiler/Fable/issues/2286 or similar issues.

plugins/Fable.React.Plugins.fs
```diff
@@ -110,8 +113,7 @@ type ReactComponentAttribute(exportDefault: bool) =
                     let getterKind = Fable.ByKey(Fable.ExprKey(AstUtil.makeStrConst arg.DisplayName))
                     let getter = Fable.Get(Fable.IdentExpr propsArg, getterKind, Fable.Any, None)
                     (arg, getter)::bindings)
-                |> List.rev
-                |> fun bindings -> Fable.Let(bindings, decl.Body)
+                |> List.fold (fun body (k,v) -> Fable.Let(k, v, body)) decl.Body
```
alfonsogarciacaro, line 116: At the end, I've separated `Expr.Let` and `Expr.LetRec` as in F# AST so if you have a list of bindings you need to fold them (in reverted order).

plugins/Fable.React.Plugins.fsproj
```diff
@@ -8,6 +8,6 @@
     <Compile Include="Fable.React.Plugins.fs" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Fable.AST" Version="3.0.0-nagareyama-beta-002" />
+    <PackageReference Include="Fable.AST" Version="3.0.0-nagareyama-rc-001" />
```
alfonsogarciacaro, line 11: Please also remember to update Fable.AST reference.

alfonsogarciacaro:
Closing as we're recommending `Feliz.ReactComponent` instead.

kerams:
@alfonsogarciacaro, has the development workflow changed? My plugin does not seem to be loaded. `--exclude` and `ScanForPlugins` are present, yet the annotated entities don't seem to trigger the `compiler.LogWarning` in `Transform` and `TransformCall`. The project compiles as if the attributes were not present.

alfonsogarciacaro:
@kerams Did you add the `ScanForPlugins` assembly attribute? See [this](https://github.com/fable-compiler/fable-react/pull/207#discussion_r530607987). Do you see a message like "Loaded plugin" when starting compilation?

kerams:
Yes, as far as I can tell I followed the instructions to the letter. I only see loaded messages for Feliz plugins.

kerams:
@alfonsogarciacaro, damn, I've just noticed `Also, please put the plugins directly in a namespace and not inside a module.`. That was the problem. Thank you.

alfonsogarciacaro:
Ah, yes. I was planning to add plugins for classes too but I haven't done it yet because the need for it hadn't appeared yet 😅

chkn:
> Ah, yes. I was planning to add plugins for classes too but I haven't done it yet because the need for it hadn't appeared yet 😅 

@alfonsogarciacaro I might have a need for this coming up. Has there been any other thought/discussion/work on this, or should I just take a stab at it and send a PR?

MangelMaxime:
@chkn I think for now, the first step is to open an issue with what you are trying to archive, plus reproduction/pseudo code if possible

Like that maintainers will be able have a look at the use case, and provide feedbacks etc.

PR #208, opened 2020-11-12
Split Fable.React.Types

alfonsogarciacaro:
The main purpose of this is to put the basic interfaces of Fable.React in a separate package so Feliz doesn't need to depend on the full Fable.React but components generate with each API can still be compatible with each other. Hopefully, removing the dependency on the full Fable.React avoids versioning issues and it should also help a bit with Fable compilation as it doesn't need to compile all Fable.React sources.

@Zaid-Ajaj @Shmew please have a look at the Fable.React.Types. I had a look at Feliz code and I think it only needs `ReactElement, IRefValue, IContext` but I added a couple more for completeness. I've seen that Feliz also calls the `fragment` helper but if we're going to remove the Fable.React dependency, it's probably to just write your own.


alfonsogarciacaro:
Unfortunately this breaks current applications, see https://github.com/fable-compiler/fable-react/issues/210#issuecomment-735351197

PR #219, opened 2021-04-29
Introduce strf

kerams:
A simple shorthand to get rid of an annoying pipe when rendering a formatted string:
`div [] [ sprintf "Hello, %s. It is %A" user DateTime.Now |> str ]`
=>
`div [] [ strf "Hello, %s. It is %A" user DateTime.Now ]`


alfonsogarciacaro:
Thanks!

PR #225, opened 2022-02-23
Add useLayoutEffect

forki:
Ass discussed in https://github.com/fable-compiler/Fable/issues/2826


alfonsogarciacaro:
I think we need to add the dummy implementation for SSR below and not sure if we need the "disposable" overload, but I can do that. Thanks!

PR #226, opened 2022-03-19
React 18 APIs

kerams:
Initial work based on https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html

I'll add xmldoc once the documentation is up.


src/Fable.ReactDom.fs
```diff
@@ -36,6 +46,13 @@ module ReactDomBindings =
     #endif
     let ReactDom: IReactDom = jsNative
 
+    #if FABLE_REPL_LIB
+    [<Global("ReactDOM")>]
+    #else
+    [<Import("*", "react-dom/client")>]
```
kerams, line 52: Should this be `*` or `default`? What about the `FABLE_REPL_LIB` bit?

kerams:
So I've tested this in an Elmish app with the following code and it appears to work fine.

```fsharp
type IReactRoot =
    abstract render: element: ReactElement -> unit

type IReactDomClient =
    abstract createRoot: container: Element -> IReactRoot

[<Import("*", "react-dom/client")>]
let ReactDomClient: IReactDomClient = jsNative

let withReactSynchronousUsing lazyView2With placeholderId (program:Elmish.Program<_,_,_,_>) =
    let root = ReactDomClient.createRoot (Browser.Dom.document.getElementById placeholderId)

    let setState model dispatch =
        root.render (lazyView2With (fun x y -> obj.ReferenceEquals(x,y)) (Program.view program) model dispatch)

    program
    |> Program.withSetState setState

let withReactSynchronous placeholderId (program:Elmish.Program<_,_,_,_>) =
    withReactSynchronousUsing lazyView2With placeholderId program

Program.mkProgram init update view
|> withReactSynchronous "elmish"
|> Program.run
```

The only minor problem is that during an HMR reload, it prints `Warning: You are calling ReactDOMClient.createRoot() on a container that has already been passed to createRoot() before. Instead, call root.render() on the existing root instead if you want to update it.` Not sure how this could be solved, but it's an issue for elmish/react anyway.

kerams:
Perhaps I could add `/// Requires React 18.` for `createRoot` and `hydrateRoot`?

src/Fable.ReactDom.fs
alfonsogarciacaro: You can skip that, we used it for the repl in Fable 2 but we haven't uploaded external libraries yet to the repl for Fable 3.

alfonsogarciacaro:
Thanks for this @kerams! Please let us know when it's ready to merge.

> Perhaps I could add /// Requires React 18. for createRoot and hydrateRoot?

Yes, that makes sense 👍 

> The only minor problem is that during an HMR reload, ...

I haven't tried yet these new APIs in React 18 so not sure either. I think at one point Elmish was using a stateful component and used `setState` to trigger re-renders, instead of calling react-dom/render every time. Maybe that would avoid this issue?

kerams:
Done. Some [new hooks and whatnot](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#new-apis-for-libraries) are coming too, but those can be added later once they are documented for the GA release and aren't needed to get React 18 up and running in Fable.

MangelMaxime:
We can also decide to include Femto metadata, so if the user use Femto he will be warned that Fable.React now needs react >= 18.

It should be ok, as React 18 is backward compatible with previous APIs if I read correctly the blog post.

kerams:
Yes, you'll keep getting console warnings though.

I could add the Femto bits to fsproj, but what if someone wants to stay on React 17 and updates fable-react to latest?

MangelMaxime:
> I could add the Femto bits to fsproj, but what if someone wants to stay on React 17 and updates fable-react to latest?

In this case, they will need to manually manage the NPM packages I guess.

kerams:
Alright, added, it will start working with the stable release of v18.

MangelMaxime:
If we put the femto metadata for stable 18, we "can't" really publish a beta version of fable-react as it will be confusing to not be able to resolved the version requested.

kerams:
How do you specify 18 including a release candidate? Also what if we do a major version bump here too?

alfonsogarciacaro:
Sorry, I missed the notification this was ready 😅 It seems React 18 is out already so I guess we can just use the non-beta version for Femto metadata?

kerams:
Alright, I'll revert it :)

kerams:
@alfonsogarciacaro, @MangelMaxime, this is ready for the final review. I've added some other new stuff. The only thing missing are SSR API additions and [the 2 library hooks](https://reactjs.org/docs/hooks-reference.html#library-hooks).

alfonsogarciacaro (approved):

alfonsogarciacaro:
Thanks again for this @kerams! I will push a new release 👍

alfonsogarciacaro:
Given that this is mainly intended to be used with a new React major release I also bumped Fable.React version to 8.0. Should be available in Nuget now.

Issue #241, opened 2024-05-16
FunctionComponent.Of functions cache breaks reusable code


DunetsNM:
This is how `FunctionComponent.Of` caches passed function:

```
// Cache the render function to prevent recreating the component every time when FunctionComponent.Of
// is called inside another function (including generic values: let MyCom<'T> = ...)
let cacheKey = __callingSourceFile.Value + "#L" + (string __callingSourceLine.Value)
```

I have a reusable piece of code that uses `FunctionComponent.Of` for different cases and  they result in the same cacheKey.  So it picks a wrong function at runtime for all cases but one.

At least the `displayName` should also be part of the cacheKey (I pass a custom value anyway) or accept optional custom cacheKey (although that would reveal its implementation).  

None of the options is available which keeps me pinned to Fable.React 5.4.0 (last version before function caching was intorduced)

DunetsNM:
Actually my problem was that caller of the `Of` is also generic code. 
So it's better to include `'Props` type name  into the cache key so there's no need to expose internal caching mechanism to library users (requires `Of` to become `inline` though).

PR #242, opened 2024-05-17
FunctionComponent.Of extend func cache key with 'Props type name

DunetsNM:
User code that invokes `FunctionComponent.Of` can also be in generic context which makes caching by source FileName+LineNumber unsuitable:  one function gets cached and resolved at runtime for different runtime types which results in runtime exceptions.

I propose to add `'Props` type name to the cache key to fix that.

You still can write code were wrong function will be picked up at runtime for the same 'Props argument type (e.g. if different `render` func values are passed from the outside), but it's less likely to happen. 

Ultimate solution would be to expose the fact that function is prepared and cached to the users and let them to specify their own unique cache key - or piggyback on existing `displayName`, which is also not ideal.

### DIFF
--- a/src/Fable.React/Fable.React.fsproj
+++ b/src/Fable.React/Fable.React.fsproj
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <Version>9.4.0</Version>
-    <PackageVersion>9.4.0</PackageVersion>
+    <Version>9.3.0</Version>
+    <PackageVersion>9.3.0</PackageVersion>
     <TargetFramework>netstandard2.0</TargetFramework>
     <!-- <DefineConstants>$(DefineConstants);FABLE_COMPILER</DefineConstants> -->
     <PackageTags>fsharp;fable;javascript;f#;js;react;fable-library;fable-javascript;fable-dotnet</PackageTags>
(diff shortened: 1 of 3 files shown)


DunetsNM:
this fixes #241

src/Fable.React/Fable.React.fsproj
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <Version>9.3.0</Version>
-    <PackageVersion>9.3.0</PackageVersion>
+    <Version>9.4.0</Version>
+    <PackageVersion>9.4.0</PackageVersion>
```
MangelMaxime, line 5: Note for publish:

These lines should not be changed manually but instead a new entry should be made in the RELEASE_NOTES.md.

The build process will take care of updating the fsproj.

DunetsNM: fixed

MangelMaxime:
To avoid impacting too much the bundle size, we could extract the method body into a function and pass it the `typeof<'Props>.FullName` + all the required information as an argument.

What do you think?

DunetsNM:
@MangelMaxime I moved the fat part to Cache.GetOrAdd to avoid creating more public/inlineable impl functions

MangelMaxime:
Thank you

